PR #7, opened 2019-09-19
[RSA] New methods:

danielinux:
- .make_key()
- .encode_key()


danielinux:
I was indeed waiting for the test results

PR #9, opened 2019-09-27
Build Update

ejohnstown:
1. Use the same docker image as the wolfSSL python project.
2. Remove the absolute path to the python interpreter. They just need to be in the path. (For example, you might install old versions in /opt and add their subdirs to the path while the latest runs out of its location.)


ejohnstown:
This has the same problem as pull request #12 of wolfSSL-py has, only this accidentally works. Will update as time allows.

cconlon (approved):

PR #8, opened 2019-09-24
Native feature detection, CFFI cleanup

cconlon:
This PR includes:

- Native feature detection based on <wolfssl/options.h>
- cffi cleanup
- Ability to compile with pip against local wolfSSL library (USE_LOCAL_WOLFSSL), ie:

```
cd wolfcrypt-py
USE_LOCAL_WOLFSSL=/path/to/install pip install .
```

PR #19, opened 2021-01-20
Add support for RSA private keys in PKCS #8 format.

haydenroche5:
See ZD 11524.

### DIFF
--- a/src/wolfcrypt/ciphers.py
+++ b/src/wolfcrypt/ciphers.py
@@ -435,11 +435,18 @@ if _lib.RSA_ENABLED:
 
             if key != None:
                 key = t2b(key)
-
                 ret = _lib.wc_RsaPrivateKeyDecode(key, idx,
                                               self.native_object, len(key))
                 if ret < 0:
-                    raise WolfCryptError("Invalid key error (%d)" % ret)
+                    idx[0] = 0
+                    ret = _lib.wc_GetPkcs8TraditionalOffset(key, idx, len(key))
+                    if ret < 0:
+                        raise WolfCryptError("Invalid key error (%d)" % ret)
+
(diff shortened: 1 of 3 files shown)


tests/test_ciphers.py
```diff
@@ -293,6 +337,22 @@ def test_rsa_encrypt_decrypt(rsa_private, rsa_public):
         assert 1024 / 8 == len(ciphertext) == rsa_private.output_size
         assert plaintext == rsa_private.decrypt(ciphertext)
 
+    def test_rsa_pkcs8_encrypt_decrypt(rsa_private_pkcs8, rsa_public):
+        plaintext = t2b("Everyone gets Friday off.")
+
+        # normal usage, encrypt with public, decrypt with pirate
```
cconlon, line 343: minor typo, "pirate" should be "private"

cconlon (changes requested):

src/wolfcrypt/ciphers.py
```diff
@@ -435,11 +435,18 @@ def __init__(self, key = None):  # pylint: disable=super-init-not-called
 
             if key != None:
                 key = t2b(key)
-
                 ret = _lib.wc_RsaPrivateKeyDecode(key, idx,
                                               self.native_object, len(key))
                 if ret < 0:
-                    raise WolfCryptError("Invalid key error (%d)" % ret)
+                    idx[0] = 0
```
haydenroche5, line 441: Per some feedback from the ZD ticket, I opted to not add a pkcs8 parameter and instead just try to decode the key in pkcs8 format if the traditional decoding fails. Makes the interface a bit cleaner.

cconlon: Sounds good, thanks

cconlon (approved):

PR #21, opened 2021-02-24
remove python 2.7 from make/osx scripts

cconlon:
This PR removes Python 2.7 from the make/osx scripts.  Python 2.7 reached End of Life on January 1, 2020.  pip 21.0 dropped support for Python 2.7 in January 2021.

### DIFF
--- a/make/osx/build_wheels.sh
+++ b/make/osx/build_wheels.sh
@@ -1,7 +1,7 @@
 set -e
 set +x
 
-for VER in 2.7 3.7 3.8; do
+for VER in 3.7 3.8; do
     PIP="pip${VER}"
     PYTHON="python${VER}"
     VENV="venv_${VER}"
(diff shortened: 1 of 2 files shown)


haydenroche5:
`Traceback (most recent call last):
  File "<stdin>", line 24226, in <module>
  File "<stdin>", line 199, in main
  File "<stdin>", line 82, in bootstrap
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmpMIX_9g/pip.zip/pip/_internal/cli/main.py", line 60
    sys.stderr.write(f"ERROR: {exc}")
                                   ^
SyntaxError: invalid syntax`
This was the error on my PR, which does indeed look a lot like an issue of Python 2 vs. 3. F-strings weren't added until Python 3.6: https://www.python.org/dev/peps/pep-0498/. Hopefully this fixes the issue! Thanks Chris.

PR #24, opened 2021-09-16
Make wolfcrypt-py work with FIPS ready and other improvements.

haydenroche5:
- Detect ECC timing resistance and call wc_ecc_set_rng where appropriate.
- Detect FIPS version and use that information to figure out how to map hash
enum values (see _TYPE_SHA and friends).
- Don't call wc_HmacSetKey in the _Hmac constructor if the key passed to _init
is length 0. This can happen, for example, when the _Hmac object is being
copied. The copy operation copies over the raw memory from the underlying C
object, so it's not important that we call wc_HmacSetKey in this case.
- Removed a unit test that expected importing an ECC public key from a private
key to fail. This does fail in the default wolfSSL version for wolfcrypt-py,
v4.1.0-stable, but we added the feature to be able to import public from
private with wolfSSL PR #2916. As a result, this test fails with v4.8.1-stable.
We should upgrade wolfcrypt-py's default wolfSSL version (and the wolfcrypt-py
version itself) in the near future.
- The array slicing in test_key_encoding was wrong in many places. This likely
stemmed from the author thinking slices were inclusive, but that's only true
for the first element of the slice (e.g. [0:31] is elements 0-30 inclusive, not
elements 0-31 inclusive). This was uncovered by testing with FIPS ready, which
adds -DWOLFSSL_VALIDATE_ECC_IMPORT, causing us to check ECC keys with
wc_ecc_check_key. wc_ecc_check_key kept saying, "hey, that point's not on the
curve." The array slicing problem was the culprit.
- Fixed tests that were doing HMAC with a key less than HMAC_FIPS_MIN_KEY.

### DIFF
--- a/src/wolfcrypt/_build_ffi.py
+++ b/src/wolfcrypt/_build_ffi.py
@@ -20,6 +20,7 @@
 
 import os
 import sys
+import re
 from distutils.util import get_platform
 from cffi import FFI
 from wolfcrypt import __wolfssl_version__ as version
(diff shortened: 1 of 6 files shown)


danielinux (approved):

danielinux:
Thanks @haydenroche5 !

PR #26, opened 2021-12-16
Add a pem_to_der function and support for PEM RSA keys.

haydenroche5:


src/wolfcrypt/asn.py
```diff
@@ -0,0 +1,41 @@
...
+
+# pylint: disable=no-member,no-name-in-module
+
+from wolfcrypt._ffi import ffi as _ffi
+from wolfcrypt._ffi import lib as _lib
+
+from wolfcrypt.exceptions import WolfCryptError
+
+if _lib.ASN_ENABLED:
+    def pem_to_der(file, pem_type):
```
danielinux, line 29: I'm not sure why it should get a file as argument, and not a buffer. It's a bit confusing to me since wc_PemToDer works on buffers.

Would it perhaps make more sense that the caller invokes `open`/`read` on the file, retrieves the pem buffer, then calls `pem_to_der()` on it?

danielinux (changes requested):
totally in favor of adding a ASN module with utility functions for conversion.

Two things:
- wolfcrypt.asn.pem_to_der should use a buffer (not a file name) as input
- would be nice to have wolfcrypt.asn.der_to_pem doing the inverse operation

src/wolfcrypt/asn.py
haydenroche5: Good feedback. I agree.

danielinux (approved):

danielinux:
Nice feature addition! Thanks @haydenroche5

PR #34, opened 2022-01-31
Fix sdist and packaging

LinuxJedi:
wolfcrypt was not being packaged in the sdist and the build system did
not work if the source wasn't a git checkout. Both fixed.


danielinux (approved):

danielinux:
this fixes the build everywhere for me. Thanks @LinuxJedi !

PR #36, opened 2022-02-03
Add RSA OAEP and PSS padding

LinuxJedi:
Functions added:
* `encrypt_oaep(self, plaintext, hash_type, mgf, label)`
* `decrypt_oaep(self, ciphertext, hash_type, mgf, label)`
* `sign_pss(self, plaintext, hash_type, mgf)`
* `verify_pss(self, plaintext, signature, hash_type, mgf)`

Constants added:
* MGF1SHA224
* MGF1SHA256
* MGF1SHA384
* MGF1SHA512
* HASH_TYPE_NONE
* HASH_TYPE_MD2
* HASH_TYPE_MD4
* HASH_TYPE_MD5
* HASH_TYPE_SHA
* HASH_TYPE_SHA224
* HASH_TYPE_SHA256
* HASH_TYPE_SHA384
* HASH_TYPE_SHA512
* HASH_TYPE_MD5_SHA
* HASH_TYPE_SHA3_224
* HASH_TYPE_SHA3_256
* HASH_TYPE_SHA3_384
* HASH_TYPE_SHA3_512
* HASH_TYPE_BLAKE2B
* HASH_TYPE_BLAKE2S


wolfcrypt/_build_ffi.py
```diff
@@ -381,12 +381,22 @@ def generate_libwolfssl():
                             RsaKey* key);
     int wc_RsaPublicEncrypt(const byte*, word32, byte*, word32,
                             RsaKey*, WC_RNG*);
-
+    int wc_RsaPublicEncrypt_ex(const byte* in, word32 inLen, byte* out,
```
haydenroche5, line 384: These new additions have `#ifdef` gates in wolfSSL. This is no problem in the case where wolfcrypt-py builds wolfSSL itself (i.e. using _build_wolfssl.py), but it could be problematic when doing feature detection (i.e. using `USE_LOCAL_WOLFSSL`). For example, the PSS stuff is gated on `#ifdef WC_RSA_PSS`. I think we should add some logic to the feature detection stuff in this file to account for that. The *_ex functions also have gates that need to be accounted for.

wolfcrypt/ciphers.py
```diff
@@ -82,6 +82,39 @@
 ECC_BRAINPOOLP384R1 = 26
 ECC_BRAINPOOLP512R1 = 27
 
+RSA_PKCSV15_PAD = 0
```
haydenroche5, line 85: If I recall correctly, macro constants can be pulled from the underlying library directly. This definitely works for enums (see `static const long PRIVATEKEY_TYPE;` in _build_ffi.py for an example). If that works, we should do that, IMO, instead of repeating ourselves with the same constant definitions here.

haydenroche5 (changes requested):
Looks good overall! Just a few comments.

wolfcrypt/_build_ffi.py
LinuxJedi: Good point (although the current gating is a bit broken). I'll add it.

wolfcrypt/ciphers.py
LinuxJedi: Ah! That wasn't done elsewhere so figured I'd follow the pattern. I'll fix it.

wolfcrypt/_build_ffi.py
```diff
@@ -381,11 +385,53 @@ def generate_libwolfssl():
...
+        static const int WC_RSA_NO_PAD;
+
+        static const int WC_MGF1NONE;
+        static const int WC_MGF1SHA1;
+        static const int WC_MGF1SHA224;
+        static const int WC_MGF1SHA256;
+        static const int WC_MGF1SHA384;
+        static const int WC_MGF1SHA512;
+
+        static const int WC_HASH_TYPE_NONE;
```
haydenroche5, line 409: Do all of the constants here need to be gated on `RSA_PSS_ENABLED`? I think many of these can be added unconditionally.

tests/test_ciphers.py
```diff
@@ -415,6 +432,22 @@ def test_rsa_sign_verify(rsa_private, rsa_public):
         assert 1024 / 8 == len(signature) == rsa_private.output_size
         assert plaintext == rsa_private.verify(signature)
 
+    def test_rsa_pss_sign_verify(rsa_private, rsa_public):
```
haydenroche5, line 435: Should we gate these PSS tests on `_lib.RSA_PSS_ENABLED`?

haydenroche5 (changes requested):

wolfcrypt/_build_ffi.py
LinuxJedi: Technically they all could. I'll move them out.

wolfcrypt/ciphers.py
```diff
@@ -82,6 +82,31 @@
 ECC_BRAINPOOLP384R1 = 26
 ECC_BRAINPOOLP512R1 = 27
 
+MGF1NONE = _lib.WC_MGF1NONE
```
haydenroche5, line 85: These assignments will fail if not `RSA_ENABLED`, as it stands now.

haydenroche5 (changes requested):

haydenroche5 (approved):

PR #40, opened 2022-02-25
Simplify OAEP and PSS

LinuxJedi:
Makes things a little bit more like similar APIs.

* Hash type is now set in constructor.
* MGF is set automtically or manually with `set_mgf()`
* Label defaults to empty


LinuxJedi:
Draft PR until #39 is merged

danielinux (approved):

PR #49, opened 2022-09-15
Make several improvements to the CFFI build process.

haydenroche5:
- Move _build_ffi.py to scripts/build_ffi.py. Do as little in the global namespace of the script as possible by putting functionality into functions, including a new main function.
- Improve feature detection by allowing user_settings.h in addition to options.h. This is particularly useful on Windows.
- Add the windows directory. This contains subdirectories non_fips and fips_ready, which each have a user_settings.h. The non_fips user_settings.h is used by the build script to build on Windows with CMake, when not using USE_LOCAL_WOLFSSL. The fips_ready user_settings.h is an example user_settings.h that works with our FIPS Ready code and wolfcrypt-py.
- Improve Windows support. This commit modifies the way we build wolfSSL for wolfcrypt-py on Windows. Instead of using CMake directly, we use it with user_settings.h. This is typically how Windows users are building wolfSSL, and it gets around some shortcoming of wolfSSL's CMake support (e.g. not all required features are supported). Additionally, USE_LOCAL_WOLFSSL now works properly on Windows. When using USE_LOCAL_WOLFSSL, it should point to a wolfSSL source code directory. This works for FIPS bundles, too. Finally, for FIPS on Windows, we only support DLL builds. The CFFI library (.pyd extension) requires this DLL to work. build_ffi.py detects this scenario and copies the .dll into the wolfcrypt package directory. I've modified setup.py to distribute the .dll with the package.
- Add some missing FIPS gates. Testing against FIPS v2 triggered some errors. Those are fixed in this commit.


haydenroche5:
Got a build failure here: https://github.com/haydenroche5/wolfcrypt-py/actions/runs/3056836857

Will look into it tomorrow.

danielinux (approved):
no regressions in build/tests on Linux

danielinux:
Closing in favor of #53

PR #64, opened 2024-09-13
Adding ChaCha20/Poly1305 Support

jackctj117:
Added to build ffi and added to cipher.py. Not exactly sure what to add in cipher.py and same goes for the test.

### DIFF
--- a/scripts/build_ffi.py
+++ b/scripts/build_ffi.py
@@ -642,19 +642,19 @@ def build_ffi(local_wolfssl, features):
     if features["CHACHA20_POLY1305"]:
         cdef += """
         typedef struct { ...; } ChaChaPoly_Aead;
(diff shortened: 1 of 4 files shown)


jackctj117:
Added encrypt and decrypt test, didn't include stream functionality.

scripts/build_ffi.py
```diff
@@ -215,7 +215,7 @@ def make_flags(prefix, fips):
 
         flags.append("--disable-md5")
         flags.append("--disable-sha224")
-        flags.append("--disable-poly1305")
+        flags.append("--enable-poly1305")
```
dgarske, line 218: `--enable-chacha`?

dgarske (dismissed):

scripts/build_ffi.py
jackctj117: Just a few lines up it's enabled

danielinux (approved):
Tested OK

PR #77, opened 2025-11-24
build_ffi.py: Fix header parsing in feature detection logic

mjdemilliano:
`defines` was a string instead of a list of strings. This made it impossible to build a wrapper with a configuration in which AES-CBC is disabled, because `'#define NO_AES' in defines` matched also when the header contains `#define NO_AES_CBC`. With `defines` being a list the code actually does what you think it does.


lealem47 (approved):
This is a great adjustment! Thank you @mjdemilliano

PR #78, opened 2025-11-24
Move some cdef declarations for RSA to separate section

mjdemilliano:
The definitions for the following functions are not always available when wolfSSL is configured to only contain a minimum of functionality:
- wc_GetPkcs8TraditionalOffset (only with ASN, which is required for RSA)
- wc_PemToDer (only when KEYGEN is enabled)
- wc_DerToPemEx (only when KEYGEN is enabled)

This change allows for building the Python wrapper when ASN and RSA are both disabled.

### DIFF
--- a/scripts/build_ffi.py
+++ b/scripts/build_ffi.py
@@ -957,21 +957,17 @@ def build_ffi(local_wolfssl, features):
 
         word32 wc_EncodeSignature(byte* out, const byte* digest, word32 digSz,
                                   int hashOID);
+        int wc_PemToDer(const unsigned char* buff, long longSz, int type,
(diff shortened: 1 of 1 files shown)


scripts/build_ffi.py
```diff
@@ -743,6 +741,8 @@ def build_ffi(local_wolfssl, features):
         int wc_InitRsaKey(RsaKey* key, void*);
         int wc_FreeRsaKey(RsaKey* key);
 
+        int wc_GetPkcs8TraditionalOffset(byte* input, word32* inOutIdx, word32 sz);
```
lealem47: This should depend on the ASN feature, instead of the RSA feature.

lealem47: Perfect, thanks!

PR #100, opened 2026-04-11
Start adding typing information.

roberthdevries:
This is very preliminary and only provides type information for random.py and utils.py

### DIFF
--- a/wolfcrypt/_ffi/__init__.pyi
+++ b/wolfcrypt/_ffi/__init__.pyi
@@ -2,4 +2,4 @@ import _cffi_backend
 
 ffi: _cffi_backend.FFI
 
-__all__ = ["lib"]
+__all__ = ["ffi", "lib"]
--- a/wolfcrypt/_ffi/__init__.pyi
+++ b/wolfcrypt/_ffi/__init__.pyi
@@ -1,5 +1,5 @@
 import _cffi_backend
-import _ffi.lib as lib
+import wolfcrypt._ffi.lib as lib
 
 ffi: _cffi_backend.FFI
 
--- a/wolfcrypt/_ffi/__init__.pyi
+++ b/wolfcrypt/_ffi/__init__.pyi
@@ -1,4 +1,5 @@
 import _cffi_backend
+import _ffi.lib as lib
 
 ffi: _cffi_backend.FFI
 
--- a/wolfcrypt/_ffi/__init__.pyi
+++ b/wolfcrypt/_ffi/__init__.pyi
@@ -1,0 +1,5 @@
+import _cffi_backend
+
+ffi: _cffi_backend.FFI
+
+__all__ = ["lib"]
--- a/wolfcrypt/_ffi/lib.pyi
+++ b/wolfcrypt/_ffi/lib.pyi
@@ -1,0 +1,76 @@
+
+from _cffi_backend import FFI
+from typing_extensions import TypeAlias
+
+INVALID_DEVID: int
+
+AES_ENABLED: int
+AES_SIV_ENABLED: int
+AESGCM_STREAM_ENABLED: int
+ASN_ENABLED: int
+CHACHA_ENABLED: int
+CHACHA_STREAM_ENABLED: int
+CHACHA20_POLY1305_ENABLED: int
+DES3_ENABLED: int
+ECC_ENABLED: int
+ED25519_ENABLED: int
+ED448_ENABLED: int
+FIPS_ENABLED: int
+HMAC_ENABLED: int
+KEYGEN_ENABLED: int
+HKDF_ENABLED: int
+ML_DSA_ENABLED: int
+ML_KEM_ENABLED: int
+MPAPI_ENABLED: int
+PWDBASED_ENABLED: int
+RSA_ENABLED: int
+RSA_PSS_ENABLED: int
+SHA_ENABLED: int
+SHA3_ENABLED: int
+SHA256_ENABLED: int
+SHA384_ENABLED: int
+SHA512_ENABLED: int
+WC_RNG_SEED_CB_ENABLED: int
+
+FIPS_VERSION: int
+
+WC_MGF1NONE: int
+WC_MGF1SHA1: int
+WC_MGF1SHA224: int
+WC_MGF1SHA256: int
+WC_MGF1SHA384: int
+WC_MGF1SHA512: int
+
+WC_HASH_TYPE_NONE: int
+WC_HASH_TYPE_MD2: int
+WC_HASH_TYPE_MD4: int
+WC_HASH_TYPE_MD5: int
+WC_HASH_TYPE_SHA: int
+WC_HASH_TYPE_SHA224: int
+WC_HASH_TYPE_SHA256: int
+WC_HASH_TYPE_SHA384: int
+WC_HASH_TYPE_SHA512: int
+WC_HASH_TYPE_MD5_SHA: int
+WC_HASH_TYPE_SHA3_224: int
+WC_HASH_TYPE_SHA3_256: int
+WC_HASH_TYPE_SHA3_384: int
+WC_HASH_TYPE_SHA3_512: int
+WC_HASH_TYPE_BLAKE2B: int
+WC_HASH_TYPE_BLAKE2S: int
+
+WC_ML_KEM_512: int
+WC_ML_KEM_768: int
+WC_ML_KEM_1024: int
+
+WC_ML_DSA_44: int
+WC_ML_DSA_65: int
+WC_ML_DSA_87: int
+
+WC_KEYTYPE_ALL: int
+
+RNG: TypeAlias = FFI.CData
+
+def wc_InitRngNonce_ex(rng: RNG, nonce: bytes, nonce_size: int, heap: FFI.CData, device_id: int) -> int: ...
+def wc_RNG_GenerateByte(rng: RNG, buffer: FFI.CData) -> int: ...
+def wc_RNG_GenerateBlock(rng: RNG, buffer: FFI.CData, len: int) -> int: ...
+def wc_FreeRng(rng: RNG) -> None: ...
--- a/wolfcrypt/_ffi/lib.pyi
+++ b/wolfcrypt/_ffi/lib.pyi
@@ -1,6 +1,6 @@
 
 from _cffi_backend import FFI
-from typing_extensions import TypeAlias
+from typing import TypeAlias
 
 INVALID_DEVID: int
 
--- a/wolfcrypt/random.py
+++ b/wolfcrypt/random.py
@@ -22,8 +22,6 @@
 
 from __future__ import annotations
 
-from _cffi_backend import FFI
-
 from wolfcrypt._ffi import ffi as _ffi
 from wolfcrypt._ffi import lib as _lib
 
--- a/wolfcrypt/random.py
+++ b/wolfcrypt/random.py
@@ -47,7 +47,7 @@ class Random:
     def __del__(self) -> None:
         if self.native_object:
             try:
-                Random._delete(self.native_object)
+                self._delete(self.native_object)
             except AttributeError:
                 # Can occur during interpreter shutdown
                 pass
--- a/wolfcrypt/random.py
+++ b/wolfcrypt/random.py
@@ -34,7 +34,7 @@ class Random:
     """
 
     def __init__(self, nonce: __builtins__.bytes = b"", device_id: int = _lib.INVALID_DEVID) -> None:
-        self.native_object = _ffi.new("WC_RNG *")
+        self.native_object: _lib.RNG | None = _ffi.new("WC_RNG *")
 
         ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, len(nonce), _ffi.NULL, device_id)
         if ret < 0:  # pragma: no cover
--- a/wolfcrypt/random.py
+++ b/wolfcrypt/random.py
@@ -20,6 +20,10 @@
 
 # pylint: disable=no-member,no-name-in-module
 
+from __future__ import annotations
+
+from _cffi_backend import FFI
+
 from wolfcrypt._ffi import ffi as _ffi
 from wolfcrypt._ffi import lib as _lib
 
@@ -31,14 +35,10 @@ class Random:
     A Cryptographically Secure Pseudo Random Number Generator - CSPRNG
     """
 
-    def __init__(self, nonce=_ffi.NULL, device_id=_lib.INVALID_DEVID):
+    def __init__(self, nonce=b"", device_id=_lib.INVALID_DEVID) -> None:
         self.native_object = _ffi.new("WC_RNG *")
 
-        if nonce == _ffi.NULL:
-            nonce_size = 0
-        else:
-            nonce_size = len(nonce)
-        ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, nonce_size, _ffi.NULL, device_id)
+        ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, len(nonce), _ffi.NULL, device_id)
         if ret < 0:  # pragma: no cover
             self.native_object = None
             raise WolfCryptError("RNG init error (%d)" % ret)
@@ -46,32 +46,34 @@ class Random:
     # making sure _lib.wc_FreeRng outlives WC_RNG instances
     _delete = _lib.wc_FreeRng
 
-    def __del__(self):
+    def __del__(self) -> None:
         if self.native_object:
             try:
-                self._delete(self.native_object)
+                Random._delete(self.native_object)
             except AttributeError:
                 # Can occur during interpreter shutdown
                 pass
 
-    def byte(self):
+    def byte(self) -> __builtins__.bytes:
         """
         Generate and return a random byte.
         """
-        result = _ffi.new('byte[1]')
+        result = _ffi.new("byte[1]")
 
+        assert self.native_object is not None
         ret = _lib.wc_RNG_GenerateByte(self.native_object, result)
         if ret < 0:  # pragma: no cover
             raise WolfCryptError("RNG generate byte error (%d)" % ret)
 
         return _ffi.buffer(result, 1)[:]
 
-    def bytes(self, length):
+    def bytes(self, length) -> __builtins__.bytes:
         """
         Generate and return a random sequence of length bytes.
         """
-        result = _ffi.new('byte[%d]' % length)
+        result = _ffi.new("byte[%d]" % length)
 
+        assert self.native_object is not None
         ret = _lib.wc_RNG_GenerateBlock(self.native_object, result, length)
         if ret < 0:  # pragma: no cover
             raise WolfCryptError("RNG generate block error (%d)" % ret)
--- a/wolfcrypt/random.py
+++ b/wolfcrypt/random.py
@@ -33,7 +33,7 @@ class Random:
     A Cryptographically Secure Pseudo Random Number Generator - CSPRNG
     """
 
-    def __init__(self, nonce=b"", device_id=_lib.INVALID_DEVID) -> None:
+    def __init__(self, nonce: __builtins__.bytes = b"", device_id: int = _lib.INVALID_DEVID) -> None:
         self.native_object = _ffi.new("WC_RNG *")
 
         ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, len(nonce), _ffi.NULL, device_id)
@@ -65,7 +65,7 @@ class Random:
 
         return _ffi.buffer(result, 1)[:]
 
-    def bytes(self, length) -> __builtins__.bytes:
+    def bytes(self, length: int) -> __builtins__.bytes:
         """
         Generate and return a random sequence of length bytes.
         """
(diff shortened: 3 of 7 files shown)


wolfcrypt/random.py
```diff
@@ -31,47 +35,45 @@ class Random(object):
...
         if self.native_object:
             try:
-                self._delete(self.native_object)
+                Random._delete(self.native_object)
             except AttributeError:
                 # Can occur during interpreter shutdown
                 pass
 
-    def byte(self):
+    def byte(self) -> __builtins__.bytes:
```
JeremiahM37, line 55: **🔴 [High] `__builtins__.bytes` is broken as a type annotation in non-`__main__` modules**

The return type annotation `-> __builtins__.bytes` is incorrect. In Python, `__builtins__` is a `dict` (not a module) in every module except `__main__`. Since `random.py` is always imported as a module, `__builtins__` is a dict, and `__builtins__.bytes` would raise `AttributeError` (dicts don't have a `.bytes` attribute). While `from __future__ import annotations` prevents runtime evaluation (PEP 563), this annotation will: (1) cause `AttributeError` if anyone calls `typing.get_type_hints(Random.byte)`, (2) fail validation with type checkers like mypy/pyright, and (3) confuse IDE autocompletion. The built-in `bytes` type is already in scope — no qualification is needed.

```c
def byte(self) -> __builtins__.bytes:
    ...
    def bytes(self, length) -> __builtins__.bytes:
```

**Recommendation:** Replace `__builtins__.bytes` with plain `bytes` in both return type annotations. The builtin `bytes` is always available without qualification.

wolfcrypt/random.py
```diff
@@ -20,6 +20,10 @@
 
 # pylint: disable=no-member,no-name-in-module
 
+from __future__ import annotations
+
+from _cffi_backend import FFI
```
JeremiahM37: **🟠 [Medium] Unused `from _cffi_backend import FFI` import in random.py**

The import `from _cffi_backend import FFI` was added but `FFI` is never referenced anywhere in `random.py`. The annotations use `__builtins__.bytes` (which should be plain `bytes`), not `FFI`. This import will cause a linter warning (e.g., F401 unused import) and adds a hard dependency on `_cffi_backend` at import time that didn't exist before. Note that `_cffi_backend` is a C extension module; importing it directly is unusual — the codebase already imports through `wolfcrypt._ffi`.

```c
from _cffi_backend import FFI
```

**Recommendation:** Remove the `from _cffi_backend import FFI` import since it is not used.

wolfcrypt/_ffi/__init__.pyi
```diff
@@ -0,0 +1,5 @@
+import _cffi_backend
+
+ffi: _cffi_backend.FFI
+
+__all__ = ["lib"]
```
JeremiahM37: **🟠 [Medium] `__init__.pyi` stub declares `__all__ = ["lib"]` but does not declare `lib` itself**

The stub file declares `__all__ = ["lib"]` but never defines or imports `lib`. Additionally, `ffi` is declared as a module attribute (line 3) but is excluded from `__all__`, even though `random.py` imports it with `from wolfcrypt._ffi import ffi`. While `__all__` doesn't restrict direct imports, it signals the module's public API to type checkers and documentation tools. The stub should be consistent with actual module exports.

```c
import _cffi_backend

ffi: _cffi_backend.FFI

__all__ = ["lib"]
```

**Recommendation:** Either add `ffi` to `__all__` and declare `lib` as a submodule reference, or remove `__all__` from the stub if it's not meaningful.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Findings:** 6 total — 4 posted, 2 skipped

### Posted findings
- [High] **`__builtins__.bytes` is broken as a type annotation in non-`__main__` modules** — `wolfcrypt/random.py:57,70`
- [Medium] **Unused `from _cffi_backend import FFI` import in random.py** — `wolfcrypt/random.py:25`
- [Medium] **`__init__.pyi` stub declares `__all__ = ["lib"]` but does not declare `lib` itself** — `wolfcrypt/_ffi/__init__.pyi:5`
- [Low] **Docstring in `t2b()` is now inaccurate after behavioral change** — `wolfcrypt/utils.py:30-31`

### Skipped findings
- [Medium] ~~`Random._delete` in `__del__` is less shutdown-safe than `self._delete`~~
- [Info] ~~Missing newline at end of file in `__init__.pyi`~~

wolfcrypt/random.py
roberthdevries: This is a work-around solution for the existence of the `bytes` method of this class. Another solution would be to rename the method to `randbytes` similar to the Python standard library Random class. But that means that the interface changes.
You cannot have both no `__builtin__.bytes` and a `bytes` method. Choose one of the two.

wolfcrypt/random.py
roberthdevries: Removed

wolfcrypt/_ffi/__init__.pyi
roberthdevries: added ffi

JeremiahM37 (dismissed):

wolfcrypt/random.py
```diff
@@ -31,47 +33,45 @@ class Random(object):
...
     _delete = _lib.wc_FreeRng
 
-    def __del__(self):
+    def __del__(self) -> None:
         if self.native_object:
             try:
-                self._delete(self.native_object)
+                Random._delete(self.native_object)
             except AttributeError:
                 # Can occur during interpreter shutdown
```
dgarske, line 52: **🟠 [High] `Random._delete` in `__del__` can raise uncaught `NameError` during interpreter shutdown**
*🚫 BLOCK*  `bug`

The original code used `self._delete(self.native_object)`, which resolves `_delete` via instance attribute lookup (falling back to the class attribute). The PR changes this to `Random._delete(self.native_object)`. During interpreter shutdown, the module globals (including the name `Random`) may be set to `None` or deleted before all `Random` instances are garbage-collected. When `__del__` fires in this state, `Random._delete` raises `NameError: name 'Random' is not defined`, which is **not** caught by the `except AttributeError` block. The original `self._delete` pattern is robust because `self` is always valid inside `__del__`, and it is the pattern used consistently throughout `ciphers.py` (lines 671, 1021, 1367, 1557).

**Suggestion:**
```suggestion
def __del__(self) -> None:
        if self.native_object:
            try:
                self._delete(self.native_object)
            except AttributeError:
                # Can occur during interpreter shutdown
                pass
```

wolfcrypt/random.py
```diff
@@ -20,6 +20,8 @@
 
 # pylint: disable=no-member,no-name-in-module
 
+from __future__ import annotations
+
 from wolfcrypt._ffi import ffi as _ffi
```
dgarske, line 25: **🟡 [Medium] Unused import `from _cffi_backend import FFI` in `random.py`**
*💡 SUGGEST*  `style`

`FFI` is imported from `_cffi_backend` but never referenced anywhere in the file — not in runtime code, not in type annotations. With `from __future__ import annotations`, even if it were used in an annotation, the import would not be needed at runtime. This adds an unnecessary runtime dependency on the internal `_cffi_backend` module.

**Suggestion:**
```suggestion
Remove the line `from _cffi_backend import FFI`.
```

wolfcrypt/utils.py
```diff
@@ -20,16 +20,18 @@
...
 
+from __future__ import annotations
+
 from binascii import hexlify as b2h, unhexlify as h2b  # noqa: F401
 
 
-def t2b(string):
+def t2b(string: bytes | bytearray | memoryview | str) -> bytes:
     """
     Converts text to binary.
```
dgarske, line 30: **🟡 [Medium] Docstring in `t2b` is now inaccurate after behavior change**
*💡 SUGGEST*  `api`

The docstring still says "Passes through bytes, bytearray, and memoryview unchanged" but the code now wraps all three in `bytes()`: `return bytes(string)`. This changes behavior — a `bytearray` input previously returned the same `bytearray`, now it returns a new `bytes` object. While the return type annotation `-> bytes` is now correctly satisfied, the docstring is misleading. The behavioral change itself is fine (all callers pass the result to CFFI which accepts both types), but documentation should match implementation.

**Suggestion:**
```suggestion
"""
    Converts text to binary.

    Converts bytes, bytearray, and memoryview to bytes.
    Encodes str to UTF-8 bytes.
    """
```

wolfcrypt/_ffi/__init__.pyi
```diff
@@ -0,0 +1,5 @@
+import _cffi_backend
+
+ffi: _cffi_backend.FFI
+
+__all__ = ["ffi", "lib"]
```
dgarske, line 6: **🔵 [Low] Missing newline at end of `__init__.pyi`**
*🔧 NIT*  `style`

The file `__init__.pyi` is missing a trailing newline. The diff shows `\ No newline at end of file`. POSIX convention and most linters expect a trailing newline.

**Suggestion:**
```suggestion
Add a trailing newline after the last line.
```

wolfcrypt/_ffi/lib.pyi
```diff
@@ -0,0 +1,76 @@
+
+from _cffi_backend import FFI
+from typing_extensions import TypeAlias
```
dgarske: **⚪ [Info] `typing_extensions.TypeAlias` used in stub but not declared as a project dependency**
*🔧 NIT*  `convention`

The stub file imports `TypeAlias` from `typing_extensions`, but `typing_extensions` is not listed in `install_requires` in `setup.py` (only `cffi>=1.0.0` is). Since `.pyi` files are only consumed by type checkers and not at runtime, this is not a runtime issue. However, `TypeAlias` is available in `typing` since Python 3.10 — using `typing.TypeAlias` would avoid the extra dependency for type-checking environments.

**Suggestion:**
```suggestion
from typing import TypeAlias
```

dgarske (changes requested):
## 🐺 Skoll Code Review

**Findings:** 6 total — 6 posted, 1 skipped

### Posted findings
- [High] **`__builtins__.bytes` is invalid as a return type annotation in non-`__main__` modules** — `wolfcrypt/random.py:57,70`
- [High] **`Random._delete` in `__del__` can raise uncaught `NameError` during interpreter shutdown** — `wolfcrypt/random.py:52`
- [Medium] **Unused import `from _cffi_backend import FFI` in `random.py`** — `wolfcrypt/random.py:25`
- [Medium] **Docstring in `t2b` is now inaccurate after behavior change** — `wolfcrypt/utils.py:30-32`
- [Low] **Missing newline at end of `__init__.pyi`** — `wolfcrypt/_ffi/__init__.pyi:5`
- [Info] **`typing_extensions.TypeAlias` used in stub but not declared as a project dependency** — `wolfcrypt/_ffi/lib.pyi:3`

<details><summary>Skipped findings</summary>

- [High] `__builtins__.bytes` is invalid as a return type annotation in non-`__main__` modules

</details>

*Review generated by Skoll via openclaw*

wolfcrypt/random.py
roberthdevries: The original line with `FFI` has already been removed.

wolfcrypt/_ffi/__init__.pyi
roberthdevries: Fixed.

wolfcrypt/_ffi/lib.pyi
roberthdevries: Fixed.

wolfcrypt/utils.py
roberthdevries: Fixed.

wolfcrypt/random.py
roberthdevries: Reverted the change

dgarske (dismissed):

JeremiahM37 (dismissed):

wolfcrypt/_ffi/__init__.pyi
```diff
@@ -0,0 +1,6 @@
+import _cffi_backend
+import _ffi.lib as lib
```
JeremiahM37: **🔴 [High] Stub imports non-existent top-level `_ffi` package**

`import _ffi.lib as lib` is an absolute import that tells type checkers to look for a top-level package named `_ffi`. The actual package path is `wolfcrypt._ffi.lib`, so type checkers (mypy, pyright, etc.) cannot resolve this import and the `lib` re-export in `__all__` becomes an unknown symbol. Runtime callers in `wolfcrypt/random.py` use `from wolfcrypt._ffi import lib as _lib`, so the stub does not reflect the real module graph.

```c
import _cffi_backend
import _ffi.lib as lib

ffi: _cffi_backend.FFI

__all__ = ["ffi", "lib"]
```

**Recommendation:** Use a relative import (`from . import lib as lib`) or the fully-qualified path `wolfcrypt._ffi.lib` so type checkers locate the sibling stub and `lib` in `__all__` resolves.

wolfcrypt/random.py
```diff
@@ -31,47 +33,45 @@ class Random:
...
-    def __init__(self, nonce=_ffi.NULL, device_id=_lib.INVALID_DEVID):
+    def __init__(self, nonce: __builtins__.bytes = b"", device_id: int = _lib.INVALID_DEVID) -> None:
         self.native_object = _ffi.new("WC_RNG *")
 
-        if nonce == _ffi.NULL:
-            nonce_size = 0
-        else:
-            nonce_size = len(nonce)
-        ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, nonce_size, _ffi.NULL, device_id)
+        ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, len(nonce), _ffi.NULL, device_id)
```
JeremiahM37, line 39: **🟠 [Medium] Default path now passes non-NULL pointer to `wc_InitRngNonce_ex`**

Previously, when the caller omitted `nonce`, the code explicitly passed `_ffi.NULL` with `nonce_size = 0` to `wc_InitRngNonce_ex`. The new code always passes the Python `nonce` argument; with the default `b""` cffi will marshal an empty bytes object to a valid, non-NULL `byte*` pointing at a zero-length buffer. wolfCrypt's `wc_InitRngNonce_ex` is typically written as `if (nonce != NULL && nonceSz > 0)`, in which case the behavior is equivalent — but some build-time paths check `nonce == NULL` as the "no nonce" sentinel. Since the previous explicit NULL branch was removed without verifying wolfCrypt's internal checks, this is a silent ABI-level behavior change for all default-constructed `Random()` instances (and there are many in `ciphers.py`).

```c
-        if nonce == _ffi.NULL:
-            nonce_size = 0
-        else:
-            nonce_size = len(nonce)
-        ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, nonce_size, _ffi.NULL, device_id)
+        ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, len(nonce), _ffi.NULL, device_id)
```

**Recommendation:** Preserve the original NULL-for-empty semantics explicitly, or confirm (and comment) that every supported wolfCrypt build treats a non-NULL, zero-length `nonce` identically to NULL.

wolfcrypt/random.py
```diff
@@ -31,47 +33,45 @@ class Random:
     A Cryptographically Secure Pseudo Random Number Generator - CSPRNG
     """
 
-    def __init__(self, nonce=_ffi.NULL, device_id=_lib.INVALID_DEVID):
+    def __init__(self, nonce: __builtins__.bytes = b"", device_id: int = _lib.INVALID_DEVID) -> None:
```
JeremiahM37, line 36: **🟠 [Medium] Public API break: `Random(nonce=_ffi.NULL)` no longer works**

The previous default value was `nonce=_ffi.NULL`, so any downstream user who mirrored the documented default and passed `_ffi.NULL` explicitly (or passed `None`) will now hit `TypeError: object of type 'CData' has no len()` on the unconditional `len(nonce)` call. The type annotation `bytes` silently narrows the accepted domain. This is a backward-incompatible change for a public class.

```c
def __init__(self, nonce: __builtins__.bytes = b"", device_id: int = _lib.INVALID_DEVID) -> None:
    ...
    ret = _lib.wc_InitRngNonce_ex(self.native_object, nonce, len(nonce), _ffi.NULL, device_id)
```

**Recommendation:** Either keep accepting `_ffi.NULL`/`None` for backward compatibility, or note the API break in the changelog. Reviewing `ChangeLog.rst` and bumping the version accordingly is advisable.

wolfcrypt/utils.py
```diff
@@ -20,16 +20,19 @@
...
 
-    Passes through bytes, bytearray, and memoryview unchanged.
+    Passes through bytes unchanged.
+    Objects of type bytearray or memoryview are converted to bytes.
     Encodes str to UTF-8 bytes.
     """
     if isinstance(string, (bytes, bytearray, memoryview)):
-        return string
+        return bytes(string)
     return str(string).encode("utf-8")
```
JeremiahM37: **🔵 [Low] `t2b` now copies `bytearray`/`memoryview` inputs**

The pre-existing contract was "Passes through bytes, bytearray, and memoryview unchanged." The new `return bytes(string)` branch eagerly copies any `bytearray`/`memoryview` argument into a fresh immutable `bytes`. Callers that previously relied on receiving back the original mutable `bytearray` (for example, to write into a shared buffer before passing to the wolfCrypt wrappers) will silently get a detached copy. Since `t2b` is a public helper used across tests, docs (`docs/streaming.rst`), and `hkdf.py`, this is worth calling out in the changelog even if no in-tree caller exercises the mutability.

```c
if isinstance(string, (bytes, bytearray, memoryview)):
    return bytes(string)
return str(string).encode("utf-8")
```

**Recommendation:** Confirm the behavior change is intentional and document it in `ChangeLog.rst`. If pass-through for `bytes` is still desired (to avoid the redundant no-op copy in CPython it's already a no-op, but making the branch explicit helps readers), split the isinstance check.

wolfcrypt/random.py
```diff
@@ -31,47 +33,45 @@ class Random:
...
 
-    def byte(self):
+    def byte(self) -> __builtins__.bytes:
         """
         Generate and return a random byte.
         """
-        result = _ffi.new('byte[1]')
+        result = _ffi.new("byte[1]")
 
+        assert self.native_object is not None
```
JeremiahM37, line 61: **⚪ [Info] Defensive `assert self.native_object is not None` is dead code**

`__init__` either raises `WolfCryptError` (after setting `native_object = None`) or returns a fully-initialized instance. The only other place `native_object` becomes `None` is that error path, which is unreachable from `byte()`/`bytes()` because the object was never handed back to the user. The asserts therefore only serve to narrow the Optional for a type checker; the same result can be achieved by not annotating `native_object` as Optional in the first place (e.g., initialize to a non-None value before the FFI call or restructure so the attribute type is the CData handle).

```c
assert self.native_object is not None
ret = _lib.wc_RNG_GenerateByte(self.native_object, result)
```

**Recommendation:** Consider removing the asserts once the attribute's type is tightened, since they add noise for no runtime benefit.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Findings:** 5 total — 5 posted, 0 skipped
**5 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Stub imports non-existent top-level `_ffi` package** — `wolfcrypt/_ffi/__init__.pyi:2`
- [Medium] **Default path now passes non-NULL pointer to `wc_InitRngNonce_ex`** — `wolfcrypt/random.py:36-39`
- [Medium] **Public API break: `Random(nonce=_ffi.NULL)` no longer works** — `wolfcrypt/random.py:36`
- [Low] **`t2b` now copies `bytearray`/`memoryview` inputs** — `wolfcrypt/utils.py:36-38`
- [Info] **Defensive `assert self.native_object is not None` is dead code** — `wolfcrypt/random.py:61,74`

JeremiahM37 (changes requested):

wolfcrypt/random.py
roberthdevries: Backward compatibility is not really applicable as this parameter was added after the previous release (5.8.4).
The new parameter to Random() is now mentioned in the release notes.

wolfcrypt/utils.py
roberthdevries: The function `t2b` seem to be internal only as far as I can see. In my opinion this does not seem to need a mention in `ChangeLog.rst`.
The checks for the types is split up. As an extra a `TypeError` is raised when none of the accepted types is passed.
Previously anything would happily be converted to a string (also floating point numbers, etc.)

wolfcrypt/random.py
roberthdevries: The check is there to make clear to static checkers that `self.native_object` is not `None` at that point.
The alternative is to remove it and have a `noqa` comment at that line.
Runtime performance is hardly an argument in Python as it runs slow anyway (as compared to compiled code).
The benefit of being able to run a static checker outweighs the minor added noise in my experience.

wolfcrypt/_ffi/__init__.pyi
roberthdevries: done

JeremiahM37 (approved):

dgarske (approved):

PR #101, opened 2026-04-11
Fix mutable arguments passed as default arguments.

roberthdevries:
Function defaults are evaluated once, when the function is defined.

The same mutable object is then shared across all calls to the function. If the object is modified, those modifications will persist across calls, which can lead to unexpected behavior.

### DIFF
--- a/tests/test_ciphers.py
+++ b/tests/test_ciphers.py
@@ -614,12 +614,11 @@ if _lib.ECC_ENABLED:
 
 
     def test_ecc_make_shared_secret():
-        rng = Random()
-        a = EccPrivate.make_key(32, rng=rng)
+        a = EccPrivate.make_key(32, rng=Random())
         a_pub = EccPublic()
         a_pub.import_x963(a.export_x963())
 
-        b = EccPrivate.make_key(32, rng=rng)
+        b = EccPrivate.make_key(32, rng=Random())
         b_pub = EccPublic()
         b_pub.import_x963(b.export_x963())
 
@@ -627,6 +626,13 @@ if _lib.ECC_ENABLED:
             == b.shared_secret(a) \
             == a.shared_secret(b_pub) \
             == b.shared_secret(a_pub)
+
+    def test_ecc_make_key_no_rng():
+        key = EccPrivate.make_key(32)
+        pub_key = EccPublic()
+        pub_key.import_x963(key.export_x963())
+
+        assert key.shared_secret(pub_key)
 
 if _lib.ED25519_ENABLED:
     @pytest.fixture
--- a/tests/test_ciphers.py
+++ b/tests/test_ciphers.py
@@ -25,6 +25,7 @@ import random
 import pytest
 from wolfcrypt._ffi import lib as _lib
 from wolfcrypt.ciphers import MODE_CTR, MODE_ECB, MODE_CBC, WolfCryptError
+from wolfcrypt.random import Random
 from wolfcrypt.utils import t2b, h2b
 import os
 
@@ -613,11 +614,12 @@ if _lib.ECC_ENABLED:
 
 
     def test_ecc_make_shared_secret():
-        a = EccPrivate.make_key(32)
+        rng = Random()
+        a = EccPrivate.make_key(32, rng=rng)
         a_pub = EccPublic()
         a_pub.import_x963(a.export_x963())
 
-        b = EccPrivate.make_key(32)
+        b = EccPrivate.make_key(32, rng=rng)
         b_pub = EccPublic()
         b_pub.import_x963(b.export_x963())
 
--- a/wolfcrypt/ciphers.py
+++ b/wolfcrypt/ciphers.py
@@ -2450,6 +2450,8 @@ if _lib.ML_DSA_ENABLED:
             :return: signature
             :rtype: bytes
             """
+            if rng is None:
+                rng = Random()
             msg_bytestype = t2b(message)
             in_size = self.sig_size
             signature = _ffi.new(f"byte[{in_size}]")
--- a/wolfcrypt/ciphers.py
+++ b/wolfcrypt/ciphers.py
@@ -1232,8 +1232,6 @@ if _lib.ECC_ENABLED:
 
         def __init__(self, key=None, rng=None):
             super().__init__(key)
-            if rng is None:
-                rng = Random()
             self._rng = rng
 
         @classmethod
@@ -1241,7 +1239,10 @@ if _lib.ECC_ENABLED:
             """
             Generates a new key pair of desired length **size**.
             """
+            if rng is None:
+                rng = Random()
             ecc = cls(rng=rng)
+
             ret = _lib.wc_ecc_make_key(ecc._rng.native_object, size,
                     ecc.native_object)
             if ret < 0:
--- a/wolfcrypt/ciphers.py
+++ b/wolfcrypt/ciphers.py
@@ -1229,30 +1229,29 @@ if _lib.ECC_ENABLED:
 
 
     class EccPrivate(EccPublic):
+
+        def __init__(self, key=None, rng=None):
+            super().__init__(key)
+            if rng is None:
+                rng = Random()
+            self._rng = rng
+
         @classmethod
         def make_key(cls, size, rng=None):
             """
             Generates a new key pair of desired length **size**.
             """
-            if rng is None:
-                rng = Random()
-            ecc = cls()
-
-            ret = _lib.wc_ecc_make_key(rng.native_object, size,
+            ecc = cls(rng=rng)
+            ret = _lib.wc_ecc_make_key(ecc._rng.native_object, size,
                     ecc.native_object)
             if ret < 0:
                 raise WolfCryptError("Key generation error (%d)" % ret)
 
             if _lib.ECC_TIMING_RESISTANCE_ENABLED and (not _lib.FIPS_ENABLED or
                _lib.FIPS_VERSION > 2):
-                ret = _lib.wc_ecc_set_rng(ecc.native_object, rng.native_object)
+                ret = _lib.wc_ecc_set_rng(ecc.native_object, ecc._rng.native_object)
                 if ret < 0:
                     raise WolfCryptError("Error setting ECC RNG (%d)" % ret)
-
-            # Retain the RNG so it outlives the ECC key. Even outside the
-            # timing-resistance path, wolfSSL internals may retain a pointer
-            # to the RNG; keeping the reference avoids any UAF risk.
-            ecc._rng = rng
 
             return ecc
 
@@ -2450,8 +2449,6 @@ if _lib.ML_DSA_ENABLED:
             :return: signature
             :rtype: bytes
             """
-            if rng is None:
-                rng = Random()
             msg_bytestype = t2b(message)
             in_size = self.sig_size
             signature = _ffi.new(f"byte[{in_size}]")


roberthdevries:
One test case `test_ecc_make_shared_secret` had to be fixed as it apparently relies on using the same random generator.
Reusing the one from the default argument obviously works, but is a dangerous (because implicit and invisible) dependency between the two calls to make a key. In the test case this is now explicit.

wolfcrypt/ciphers.py
```diff
@@ -1183,10 +1185,12 @@ def verify_raw(self, R, S, data):
 
     class EccPrivate(EccPublic):
         @classmethod
-        def make_key(cls, size, rng=Random()):
+        def make_key(cls, size, rng=None):
             """
             Generates a new key pair of desired length **size**.
             """
+            if rng is None:
```
JeremiahM37: **🟣 [Critical] Use-after-free: EccPrivate.make_key stores dangling RNG pointer when rng=None**

When `EccPrivate.make_key` is called without an explicit `rng` argument, the new code creates a temporary `Random()` as a local variable (line 1193). This RNG's native pointer is then stored *inside* the returned ECC key via `wc_ecc_set_rng(ecc.native_object, rng.native_object)` (line 1203) for timing resistance. When `make_key` returns, `rng` goes out of scope and becomes eligible for garbage collection. `Random.__del__` calls `wc_FreeRng`, freeing the native RNG memory. The ECC key object now holds a **dangling pointer** to freed memory. Any subsequent operation that relies on the stored RNG (e.g., `shared_secret`, signing operations when ECC timing resistance is enabled) will dereference freed memory — a use-after-free.

Before this PR, the `rng=Random()` default was evaluated once at class definition time. That single `Random` instance lived as long as the module was loaded, so the stored pointer remained valid. The mutable default was acting as a de-facto module-level singleton. This PR removes that lifetime guarantee without accounting for the `wc_ecc_set_rng` call that persists the pointer beyond the function scope.

```c
@classmethod
    def make_key(cls, size, rng=None):
        if rng is None:
            rng = Random()          # temporary, GC'd after return
        ecc = cls()
        ret = _lib.wc_ecc_make_key(rng.native_object, size, ecc.native_object)
        ...
        if _lib.ECC_TIMING_RESISTANCE_ENABLED and ...:
            ret = _lib.wc_ecc_set_rng(ecc.native_object, rng.native_object)
            # ^^^ stores raw C pointer to rng inside ecc
        return ecc
        # rng goes out of scope -> __del__ -> wc_FreeRng -> dangling pointer in ecc
```

**Recommendation:** Pin the `Random` instance to the `EccPrivate` object (`ecc._rng = rng`) after calling `wc_ecc_set_rng`, so the Python object's lifetime is tied to the key and the native RNG pointer remains valid. This is needed only in `EccPrivate.make_key` because it is the only changed function that stores the RNG pointer past the function scope.

tests/test_ciphers.py
```diff
@@ -613,11 +614,12 @@ def test_ecc_sign_verify_raw(ecc_private, ecc_public):
 
 
     def test_ecc_make_shared_secret():
```
JeremiahM37, line 616: **🟠 [Medium] Test does not cover the default rng=None path for EccPrivate.make_key**

The updated test now passes an explicit `rng=rng` to both `EccPrivate.make_key` calls. This means the default `rng=None` path (which is the one with the dangling pointer bug) is **never tested** with subsequent key usage. The only other call site (`EccPrivate.make_key(1024)` at line 520) tests an error path and doesn't exercise the timing resistance code. Adding a test that calls `make_key` without an explicit `rng` and then uses the key (e.g., `shared_secret` or `sign`) would catch the use-after-free.

```c
def test_ecc_make_shared_secret():
    rng = Random()
    a = EccPrivate.make_key(32, rng=rng)   # explicit rng, safe
    ...
    b = EccPrivate.make_key(32, rng=rng)   # explicit rng, safe
```

**Recommendation:** Add a test case that calls `EccPrivate.make_key()` without an explicit `rng` and then performs operations on the returned key to verify the RNG lifetime is handled correctly.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Findings:** 2 total — 2 posted, 0 skipped

### Posted findings
- [Critical] **Use-after-free: EccPrivate.make_key stores dangling RNG pointer when rng=None** — `wolfcrypt/ciphers.py:1192-1207`
- [Medium] **Test does not cover the default rng=None path for EccPrivate.make_key** — `tests/test_ciphers.py:616-629`

wolfcrypt/ciphers.py
```diff
@@ -1183,10 +1185,12 @@ def verify_raw(self, R, S, data):
...
+                rng = Random()
```
JeremiahM37: Same issue applies when a caller supplies their own `rng` — since `wc_ecc_set_rng` retains the pointer, the caller's `Random()` can also go out of scope after `make_key` returns:

```python
ecc = EccPrivate.make_key(32, rng=Random())
# Random __del__ runs, frees the WC_RNG
ecc.shared_secret(...)  # dangling pointer
```

The `ecc._rng = rng` assignment needs to be unconditional (after this `wc_ecc_set_rng` call), not just in the `if rng is None` branch.

tests/test_ciphers.py
roberthdevries: Added a test.

wolfcrypt/ciphers.py
roberthdevries: The random number generator is created if needed in the constructor of `EccPrivate` and assigned to a member variable that is kept as long as the instance exists.

wolfcrypt/ciphers.py
roberthdevries: Covered int the constructor of `EccPrivate`.

JeremiahM37 (dismissed):

wolfcrypt/ciphers.py
```diff
@@ -1182,21 +1184,27 @@ def verify_raw(self, R, S, data):
 
 
     class EccPrivate(EccPublic):
+
+        def __init__(self, key=None, rng=None):
```
dgarske, line 1233: **🟠 [High] EccPrivate.make_key: locally-created RNG freed while ECC key still holds pointer via wc_ecc_set_rng**
*🚫 BLOCK*  `bug`

When `rng=None` (the new default), `make_key` creates a local `Random()` on line 1193. On line 1203, `wc_ecc_set_rng(ecc.native_object, rng.native_object)` stores the native RNG pointer *inside* the ECC key for later use by wolfSSL (for timing-resistance operations). After `make_key` returns, the local `rng` variable has no remaining references. When Python's garbage collector reclaims it, `Random.__del__` calls `wc_FreeRng`, freeing the native `WC_RNG` memory. The ECC key now holds a dangling pointer.

In the **old code**, `rng=Random()` was evaluated once at function definition time and the single `Random` instance persisted for the lifetime of the module, so the pointer stored by `wc_ecc_set_rng` remained valid. This PR's fix for the mutable-default-argument antipattern is correct in general, but breaks the implicit lifetime guarantee that `wc_ecc_set_rng` depends on.

This is a use-after-free that can cause crashes or undefined behavior when the returned ECC key is subsequently used for operations that consult the stored RNG pointer (e.g., signing with timing resistance enabled).

**Suggestion:**
```suggestion
Store the `Random` instance on the returned `ecc` object so it stays alive as long as the key does:

    @classmethod
    def make_key(cls, size, rng=None):
        if rng is None:
            rng = Random()
        ecc = cls()

        ret = _lib.wc_ecc_make_key(rng.native_object, size,
                ecc.native_object)
        if ret < 0:
            raise WolfCryptError("Key generation error (%d)" % ret)

        if _lib.ECC_TIMING_RESISTANCE_ENABLED and (not _lib.FIPS_ENABLED or
           _lib.FIPS_VERSION > 2):
            ret = _lib.wc_ecc_set_rng(ecc.native_object, rng.native_object)
            if ret < 0:
                raise WolfCryptError("Error setting ECC RNG (%d)" % ret)
            ecc._rng = rng  # prevent GC from freeing the RNG

        return ecc
```

tests/test_ciphers.py
dgarske, line 616: **🟡 [Medium] No test exercises the default rng=None path for successful EccPrivate.make_key**
*💡 SUGGEST*  `test`

The test was updated to pass an explicit `rng=rng` to `EccPrivate.make_key`, which is good practice. However, there is now no test that exercises the default `rng=None` path for a *successful* `make_key` call. The only other call (`EccPrivate.make_key(1024)`) is expected to fail before reaching `wc_ecc_set_rng`. Adding a test that calls `EccPrivate.make_key(32)` without an `rng` argument and then uses the resulting key (e.g., for signing or shared secret) would catch the use-after-free regression described above and verify the default path works correctly.

**Suggestion:**
```suggestion
Add a test like:

    def test_ecc_make_key_default_rng():
        key = EccPrivate.make_key(32)  # uses default rng=None path
        # Force GC to expose dangling pointer if RNG not anchored
        import gc; gc.collect()
        pub = EccPublic()
        pub.import_x963(key.export_x963())
        signature = key.sign(b"test message")
        assert pub.verify(b"test message", signature)
```

dgarske (changes requested):
## 🐺 Skoll Code Review

**Findings:** 2 total — 2 posted, 0 skipped

### Posted findings
- [High] **EccPrivate.make_key: locally-created RNG freed while ECC key still holds pointer via wc_ecc_set_rng** — `wolfcrypt/ciphers.py:1188-1207`
- [Medium] **No test exercises the default rng=None path for successful EccPrivate.make_key** — `tests/test_ciphers.py:616-629`

*Review generated by Skoll via openclaw*

dgarske:
@roberthdevries please also resolve conflicts. Thanks

roberthdevries:
Conflicts are resolved.

wolfcrypt/ciphers.py
```diff
@@ -1229,31 +1229,30 @@ def verify_raw(self, R, S, data):
...
+            super().__init__(key)
+            if rng is None:
+                rng = Random()
+            self._rng = rng
```
JeremiahM37, line 1235: **🟠 [Medium] EccPrivate() now unconditionally allocates a Random() even when unused**

Moving `rng = Random()` into `__init__` means every `EccPrivate(...)` construction now seeds a fresh RNG, even paths that don't need one. This affects existing callers such as `EccPrivate(vectors[EccPrivate].key)` (tests/test_ciphers.py:501, 511, 517) and `EccPrivate()` followed by `decode_key_raw` (tests/test_ciphers.py:524, 526). These paths previously did no RNG work at construction. The cost is not free — `Random.__init__` calls `wc_InitRng`, which consumes OS entropy and performs DRBG instantiation. This is a behavioral regression for the common 'load an existing key to verify/sign with an explicitly-passed rng' workflow. Consider making `_rng` lazy (only constructed on first actual use) or keeping RNG creation inside `make_key` and only storing it when provided.

```c
def __init__(self, key=None, rng=None):
    super().__init__(key)
    if rng is None:
        rng = Random()   # <-- now runs on every construction
    self._rng = rng
```

**Recommendation:** Avoid eagerly constructing a `Random()` in `__init__`. Either make it lazy or only create it in the code path (`make_key`) that actually needs it.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Overall recommendation:** `COMMENT`
**Findings:** 5 total — 1 posted, 4 skipped

### Posted findings
- [Medium] **EccPrivate() now unconditionally allocates a Random() even when unused** — `wolfcrypt/ciphers.py:1233-1237`

### Skipped findings
- [Low] ~~PR description does not match the change — no mutable default was fixed~~
- [Low] ~~sign() and sign_raw() don't reuse the newly-stored self._rng~~
- [Low] ~~Unnecessary churn: explicit rng=Random() added to existing shared-secret test~~
- [Low] ~~Import of Random in tests is test-only but placed at module top with no guard~~

JeremiahM37 (changes requested):

wolfcrypt/ciphers.py
roberthdevries: Fixed. Only create random generator for `make_key()`.

JeremiahM37 (approved):

dgarske (approved):

PR #103, opened 2026-04-11
Add support for wcGetErrorString() including error codes.

roberthdevries:
It was already used, but not compiled in.
A unit test is added.

As wcGetErrorString() always returns a string the error handling in class _Hmac  has been removed.

### DIFF
--- a/tests/test_error_string.py
+++ b/tests/test_error_string.py
@@ -20,11 +20,12 @@
 
 import pytest
 
-from wolfcrypt._ffi import ffi as _ffi
 from wolfcrypt._ffi import lib as _lib
+from wolfcrypt.exceptions import error_string
 
 
 @pytest.mark.skipif(not _lib.ERROR_STRINGS_ENABLED, reason="wc_GetErrorString not enabled")
-@pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E, 0))
+@pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E))
 def test_error_string(err):
-    print(_ffi.string(_lib.wc_GetErrorString(err)).decode())
+    # Error code 0 is an invalid error code (it means success)
+    assert error_string(err) != error_string(0)
--- a/tests/test_error_string.py
+++ b/tests/test_error_string.py
@@ -24,8 +24,9 @@ from wolfcrypt._ffi import lib as _lib
 from wolfcrypt.exceptions import error_string
 
 
-@pytest.mark.skipif(not _lib.ERROR_STRINGS_ENABLED, reason="wc_GetErrorString not enabled")
-@pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E))
-def test_error_string(err):
-    # Error code 0 is an invalid error code (it means success)
-    assert error_string(err) != error_string(0)
+if _lib.ERROR_STRINGS_ENABLED:
+
+    @pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E))
+    def test_error_string(err):
+        # Error code 0 is an invalid error code (it means success)
+        assert error_string(err) != error_string(0)
--- a/tests/test_error_string.py
+++ b/tests/test_error_string.py
@@ -1,0 +1,30 @@
+# test_error_string.py
+#
+# Copyright (C) 2026 wolfSSL Inc.
+#
+# This file is part of wolfSSL. (formerly known as CyaSSL)
+#
+# wolfSSL is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# wolfSSL is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+
+import pytest
+
+from wolfcrypt._ffi import ffi as _ffi
+from wolfcrypt._ffi import lib as _lib
+
+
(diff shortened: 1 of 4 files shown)


JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`

### Posted findings
- [High] **wc_GetErrorString called unconditionally but only available when ERROR_STRINGS enabled** — `wolfcrypt/hashes.py:321`
- [High] **Test parametrize references constants unavailable when ERROR_STRINGS is disabled** — `tests/test_error_string.py:28`
- [Medium] **Test has no assertions — can never fail** — `tests/test_error_string.py:30`

wolfcrypt/hashes.py
```diff
@@ -318,11 +318,7 @@ def _init(self, hmac, key):
             if len(key) > 0:
                 ret = _lib.wc_HmacSetKey(self._native_object, hmac, key, len(key))
                 if ret < 0:
-                    err_str = "no error description found"
-                    try:
-                        err_str = _ffi.string(_lib.wc_GetErrorString(ret)).decode()
-                    except:
-                        pass
+                    err_str = _ffi.string(_lib.wc_GetErrorString(ret)).decode()
```
JeremiahM37: **🔴 [High] wc_GetErrorString called unconditionally but only available when ERROR_STRINGS enabled**

The PR removes the try/except guard around `_lib.wc_GetErrorString(ret)` in `_Hmac._init`, replacing it with an unconditional call. However, `wc_GetErrorString` is only added to the CFFI cdef when `features["ERROR_STRINGS"]` is true (build_ffi.py line 554). If a user links against a wolfSSL library compiled with `NO_ERROR_STRINGS` (e.g. via `--local-wolfssl`), the function will not exist in `_lib` and this line will raise `AttributeError` at runtime. The old code had a try/except specifically to handle this case gracefully. This is a regression.

```c
ret = _lib.wc_HmacSetKey(self._native_object, hmac, key, len(key))
                if ret < 0:
                    err_str = _ffi.string(_lib.wc_GetErrorString(ret)).decode()
                    raise WolfCryptError("wc_HmacSetKey returned {}: {}".format(ret, err_str))
```

**Recommendation:** Guard the `wc_GetErrorString` call with a check on `_lib.ERROR_STRINGS_ENABLED`, falling back to a generic message when error strings are not compiled in.

tests/test_error_string.py
```diff
@@ -0,0 +1,30 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+
+import pytest
+
+from wolfcrypt._ffi import ffi as _ffi
+from wolfcrypt._ffi import lib as _lib
+
+
+@pytest.mark.skipif(not _lib.ERROR_STRINGS_ENABLED, reason="wc_GetErrorString not enabled")
+@pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E, 0))
```
JeremiahM37: **🔴 [High] Test parametrize references constants unavailable when ERROR_STRINGS is disabled**

The `@pytest.mark.parametrize` decorator on line 28 references `_lib.WC_FAILURE`, `_lib.KEY_EXHAUSTED_E`, `_lib.NO_PASSWORD`, and `_lib.INTERRUPTED_E`. These constants are only defined in the CFFI cdef when `features["ERROR_STRINGS"]` is true (build_ffi.py line 554-815). When `ERROR_STRINGS` is disabled, these attributes do not exist on `_lib`, causing an `AttributeError` at module import time during pytest collection. The `skipif` decorator on line 27 cannot prevent this because `@pytest.mark.parametrize` evaluates its arguments eagerly at import time, before any skip logic runs. This will crash test collection for the entire module.

```c
@pytest.mark.skipif(not _lib.ERROR_STRINGS_ENABLED, reason="wc_GetErrorString not enabled")
@pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E, 0))
def test_error_string(err):
    print(_ffi.string(_lib.wc_GetErrorString(err)).decode())
```

**Recommendation:** Wrap the entire test function (including the parametrize decorator) inside an `if _lib.ERROR_STRINGS_ENABLED:` block, matching the pattern used elsewhere in the codebase (e.g., `if _lib.SHA_ENABLED:`, `if _lib.HMAC_ENABLED:`). This avoids accessing undefined CFFI constants at import time.

tests/test_error_string.py
```diff
@@ -0,0 +1,30 @@
...
+def test_error_string(err):
+    print(_ffi.string(_lib.wc_GetErrorString(err)).decode())
```
JeremiahM37: **🟠 [Medium] Test has no assertions — can never fail**

The test body only calls `print()` and never asserts anything. It will pass regardless of what `wc_GetErrorString` returns — an empty string, garbage, or any unexpected value would all pass silently. For error codes like `WC_FAILURE` or `KEY_EXHAUSTED_E`, the test should at minimum verify the returned string is non-empty and doesn't match a generic "unknown error" fallback. For the `0` (success) case, it could verify a known expected string.

```c
def test_error_string(err):
    print(_ffi.string(_lib.wc_GetErrorString(err)).decode())
```

**Recommendation:** Add assertions to verify the returned string is non-empty. For known error codes, consider asserting the string does not equal the unknown-error fallback message.

tests/test_error_string.py
roberthdevries: This is covered by the `@pytest.mark.skipif` directive on line 27.

wolfcrypt/hashes.py
roberthdevries: This call is moved into a dedicated function in exceptions.py. That function only uses the function when the feature is enabled. At this point the difference is invisible.

tests/test_error_string.py
roberthdevries: The test code is now changed to assert that the returned error string is not the same as the one for an error code representing success.

tests/test_error_string.py
JeremiahM37: @roberthdevries skipif doesn't help here — the parametrize tuple `(_lib.WC_FAILURE, ...)` is evaluated at import time, before pytest checks skipif. On a build without ERROR_STRINGS, pytest collection crashes with AttributeError before it ever sees the skipif. Same 3-line `if _lib.ERROR_STRINGS_ENABLED:` wrap used for `SHA_ENABLED`, `HMAC_ENABLED`, etc. fixes it:

```python
if _lib.ERROR_STRINGS_ENABLED:
    @pytest.mark.parametrize("err", (_lib.WC_FAILURE, _lib.KEY_EXHAUSTED_E, _lib.NO_PASSWORD, _lib.INTERRUPTED_E))
    def test_error_string(err):
        # Error code 0 is an invalid error code (it means success)
        assert error_string(err) != error_string(0)
```

JeremiahM37 (changes requested):

tests/test_error_string.py
roberthdevries: Fixed.

JeremiahM37 (approved):

dgarske (approved):

PR #109, opened 2026-04-22
Wrap _delete/_copy class attrs in staticmethod so self isn't bound as an extra arg.

### LINKED ISSUE #108: `_delete` class variable used in `Random` class doesn't work as intended
When calling `self._delete(self.native_object)` where `self._delete` is set to `_lib.wc_FreeRng`, the parameters passed to `wc_FreeRng` are `self` *and* `self.native_object`, instead of only `self.native_object`.

If you run the attached Python script, the function `some_delete` will print 2 parameters instead of 1.
One fix would be to call `Random._delete` or use `_lib.wc_FreeRng` directly.

[extra_self_param.py](https://github.com/user-attachments/files/26911218/extra_self_param.py)

kareem-wolfssl:
Fixes #108.

### DIFF
--- a/tests/test_delete_descriptor_binding.py
+++ b/tests/test_delete_descriptor_binding.py
@@ -1,0 +1,185 @@
+# test_delete_descriptor_binding.py
+#
+# Copyright (C) 2026 wolfSSL Inc.
+#
+# This file is part of wolfSSL. (formerly known as CyaSSL)
+#
+# wolfSSL is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# wolfSSL is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


roberthdevries:
This fix has the intended effect, but the approach using the intermediate class method `_delete` still looks strange to me. The comment `# making sure _lib.wc_FreeRng outlives WC_RNG instances`on line 44 of random.py requires some explanation. There doesn't seem to be a reason why `_lib.wc_FreeRng` would not be there at the time `__del__` of `Random` is called. Also the comment about the `AttributeError` looks extremely fishy. Again, there is no reason why `self._delete` or `self.native_object` would not exist at this point.
Can someone explain under which circumstances this could occur? My only theory at this point is that at some point in the past a buggy Python interpreter gave cause to these lines.

To following implementation of `__del__` works just as well (and is a lot simpler). (Tested on Linux with Python 3.10.12)
```
    def __del__(self) -> None:
        if self.native_object:
            _lib.wc_FreeRng(self.native_object)
```

dgarske (approved):

PR #110, opened 2026-04-24
Raise specialized WolfCryptApiError for failed API calls.

roberthdevries:
This exception takes the return value and converts that to a human-readable string (if error string support is compiled in.)

The specialized exception is derived from WolfCryptError meaning that existing users will not notice the difference except for the nicer description.

### DIFF
--- a/wolfcrypt/ciphers.py
+++ b/wolfcrypt/ciphers.py
@@ -29,8 +29,7 @@ from wolfcrypt.random import Random
 from wolfcrypt.asn import pem_to_der
 from wolfcrypt.hashes import hash_type_to_cls
 
-from wolfcrypt.exceptions import WolfCryptError
-
+from wolfcrypt.exceptions import WolfCryptError, WolfCryptApiError
 
 # key direction flags
 _ENCRYPTION = 0
@@ -196,12 +195,12 @@ class _Cipher:
             ret = self._set_key(_ENCRYPTION)
             if ret < 0:  # pragma: no cover
                 self._enc = None
-                raise WolfCryptError("Invalid key error (%d)" % ret)
+                raise WolfCryptApiError("Invalid key error", ret)
 
         result = _ffi.new("byte[%d]" % len(string))
         ret = self._encrypt(result, string)
         if ret < 0:  # pragma: no cover
-            raise WolfCryptError("Encryption error (%d)" % ret)
+            raise WolfCryptApiError("Encryption error", ret)
 
         return _ffi.buffer(result)[:]
 
@@ -229,12 +228,12 @@ class _Cipher:
             ret = self._set_key(_DECRYPTION)
             if ret < 0:  # pragma: no cover
                 self._dec = None
-                raise WolfCryptError("Invalid key error (%d)" % ret)
+                raise WolfCryptApiError("Invalid key error", ret)
 
         result = _ffi.new("byte[%d]" % len(string))
         ret = self._decrypt(result, string)
         if ret < 0:  # pragma: no cover
-            raise WolfCryptError("Decryption error (%d)" % ret)
+            raise WolfCryptApiError("Decryption error", ret)
 
         return _ffi.buffer(result)[:]
 
@@ -318,7 +317,7 @@ if _lib.AES_SIV_ENABLED:
                 associated_data, len(associated_data), nonce, len(nonce),
                 plaintext, len(plaintext), siv, ciphertext)
             if ret < 0:  # pragma: no cover
-                raise WolfCryptError("AES-SIV encryption error (%d)" % ret)
+                raise WolfCryptApiError("AES-SIV encryption error", ret)
             return _ffi.buffer(siv)[:], _ffi.buffer(ciphertext)[:]
 
         def decrypt(self, associated_data, nonce, siv, ciphertext):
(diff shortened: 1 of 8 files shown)


wolfcrypt/ciphers.py
```diff
@@ -1268,9 +1264,9 @@ def decode_key(self, key):
...
                                               self.native_object, len(key))
             if ret < 0:
-                raise WolfCryptError("Key decode error (%d)" % ret)
+                raise WolfCryptApiError("Key decode error", ret)
             if self.size <= 0:  # pragma: no cover
-                raise WolfCryptError("Key decode error (%d)" % self.size)
+                raise WolfCryptApiError("Key decode error", self.size)
             if self.max_signature_size <= 0:  # pragma: no cover
                 raise WolfCryptError(
                     "Key decode error (%d)" % self.max_signature_size)
```
JeremiahM37, line 1272: **🟠 [Medium] Inconsistent conversion in ECC `decode_key`: only one of two adjacent size checks was migrated**

In `EccPrivate.decode_key` the diff converts the `self.size <= 0` check to `WolfCryptApiError`, but the immediately adjacent `self.max_signature_size <= 0` check was left as `WolfCryptError("Key decode error (%d)" % self.max_signature_size)`. The two checks are semantically the same kind of guard (both check that a derived size attribute is positive), so leaving one in the old form and not the other is a bug-magnet: future readers will think one of them is wrong. The same pattern (one size check converted, one left behind) was missed in `EccPublic.decode_key` (lines ~1090-1095), the various `Ed25519`/`Ed448` `decode_key` paths, etc. — `self.size` was never edited there but `WolfCryptError` checks remain. Pick one and be consistent (see related finding about whether these should use WolfCryptApiError at all).

```c
if ret < 0:
                 raise WolfCryptApiError("Key decode error", ret)
             if self.size <= 0:  # pragma: no cover
                 raise WolfCryptApiError("Key decode error", self.size)
             if self.max_signature_size <= 0:  # pragma: no cover
                 raise WolfCryptError(
                     "Key decode error (%d)" % self.max_signature_size)
```

**Recommendation:** Either revert `self.size <= 0` back to WolfCryptError to match the sibling check (preferred — see API-misuse finding), or convert all the `max_signature_size <= 0` checks throughout ciphers.py to WolfCryptApiError. Don't ship a half-converted pair of adjacent guards.

wolfcrypt/ciphers.py
```diff
@@ -871,12 +869,12 @@ def make_key(cls, size, rng=None, hash_type=None):
                 ret = _lib.wc_MakeRsaKey(rsa.native_object, size, 65537,
                         rng.native_object)
                 if ret < 0:
-                    raise WolfCryptError("Key generation error (%d)" % ret)
+                    raise WolfCryptApiError("Key generation error", ret)
 
                 rsa.output_size = _lib.wc_RsaEncryptSize(rsa.native_object)
```
JeremiahM37, line 874: **🟠 [Medium] Bug exposure: `RsaPrivate.make_key` passes wrong variable as err_code**

In `RsaPrivate.make_key`, after a successful `wc_MakeRsaKey`, the code re-uses `ret` (now 0 from the success) as the error code passed to `WolfCryptApiError`:

```python
ret = _lib.wc_MakeRsaKey(...)
if ret < 0:
    raise WolfCryptApiError("Key generation error", ret)

rsa.output_size = _lib.wc_RsaEncryptSize(rsa.native_object)
rsa.size = size
if rsa.output_size <= 0:  # pragma: no cover
    raise WolfCryptApiError("Invalid key size error", ret)   # <-- ret is stale
```

The pre-existing code had the same `% ret` bug, but this PR makes the consequence worse: previously the user saw `"Invalid key size error (0)"`, now they will see `"Invalid key size error (0): no error"` (or whatever `wc_GetErrorString(0)` yields), which actively misleads debugging — the message claims success.

```c
rsa.output_size = _lib.wc_RsaEncryptSize(rsa.native_object)
                 rsa.size = size
                 if rsa.output_size <= 0:  # pragma: no cover
-                    raise WolfCryptError("Invalid key size error (%d)" % ret)
+                    raise WolfCryptApiError("Invalid key size error", ret)
```

**Recommendation:** Fix the variable: pass `rsa.output_size` (or use `WolfCryptError` since the value is a size, not a wc_* return). This is technically a pre-existing bug in the line, but the PR intensifies its user-visible damage by pumping a stale `ret=0` into `wc_GetErrorString`.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Overall recommendation:** `COMMENT`
**Findings:** 6 total — 3 posted, 3 skipped

### Posted findings
- [Medium] **Inconsistent conversion in ECC `decode_key`: only one of two adjacent size checks was migrated** — `wolfcrypt/ciphers.py:1267-1272`
- [Medium] **WolfCryptApiError fed non-error-code values (sizes, lengths) — produces misleading messages** — `wolfcrypt/ciphers.py:741,904,1269; wolfcrypt/asn.py:56-62`
- [Medium] **Bug exposure: `RsaPrivate.make_key` passes wrong variable as err_code** — `wolfcrypt/ciphers.py:874`

### Skipped findings
- [Medium] ~~No tests for the new `WolfCryptApiError` class~~
- [Low] ~~Behavioral change in `_Hmac._init`: error condition narrowed from `!= 0` to `< 0`~~
- [Low] ~~Inconsistent conversion in `asn.py`: `make_signature` and `hash_oid_from_class` left as WolfCryptError~~

wolfcrypt/ciphers.py
roberthdevries: Instead of a stale ret value, the negative value is passed to the exception.

wolfcrypt/ciphers.py
roberthdevries: Reverted the change including some minor change to use f-string instead of old-school % formatting

dgarske (approved):

PR #114, opened 2026-04-27
Add initial support for crypto callbacks.

roberthdevries:
This patch does not add support for types of callbacks. Adding support for callbacks for cipher functions is complicated by issues with cffi code generation for the `wc_CryptoInfo` structure with two layers of anonymous unions.
Uncommenting more parts of the cipher struct causes errors regarding conflicting struct sizes of other parts of the `wc_CryptoInfo` struct.
Cffi `cdef` and the compiler seem to disagree.

However as it is, callbacks work for random and hash functions.


wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,122 @@
...
+            self.device_id = device_id
+            self.ctx = _ffi.new_handle(self)
+            ret = _lib.wc_CryptoCb_RegisterDevice(device_id, _lib.py_wc_crypto_callback, self.ctx)
+            if ret < 0:  # pragma: no cover
+                raise WolfCryptError(f"CryptoCb device registration error ({ret})")
+
+        def __enter__(self) -> Self:
+            return self
+
+        def __exit__(self, exc_type, exc_value, traceback) -> None:
```
JeremiahM37: **🔴 [High] Double unregister of device on context-manager exit + GC**

Both `__exit__` and `__del__` call `self._unregister()`. When used as a context manager (the test does exactly this: `with RngCryptoCallback(10): ...`), the device is unregistered on `__exit__` and again later when the object is garbage-collected. If a second `CryptoCallback` was registered with the same device_id between exit and GC (a reasonable usage pattern, e.g., in successive tests), the GC-time unregister will rip out the live registration belonging to the new object, causing test pollution / NULL callback dispatch. There is also no flag to record "already unregistered."

```c
def __exit__(self, exc_type, exc_value, traceback) -> None:
    self._unregister()

def __del__(self) -> None:
    self._unregister()

def _unregister(self) -> None:
    _lib.wc_CryptoCb_UnRegisterDevice(self.device_id)
```

**Recommendation:** Track registration state with a `_registered` flag, set it to False after unregistering, and follow the same `try/except AttributeError` pattern used by `Random.__del__` in `wolfcrypt/random.py` to avoid errors during interpreter shutdown.

wolfcrypt/__init__.py
```diff
@@ -46,8 +46,19 @@
 if top_level_py not in ["setup.py", "build_ffi.py"]:
     from wolfcrypt._ffi import ffi as _ffi
     from wolfcrypt._ffi import lib as _lib
+    from wolfcrypt.cryptocb import CryptoCallback
     from wolfcrypt.exceptions import WolfCryptError
 
+    _lib.wolfCrypt_Init()
```
JeremiahM37: **🟠 [Medium] Return value of wolfCrypt_Init not checked**

`_lib.wolfCrypt_Init()` returns 0 on success and a negative error code on failure (e.g., FIPS power-on test failure, hardware init failure). The return value is silently discarded. Subsequent crypto calls may fail in subtle ways that are hard to diagnose if init failed. Other init points in this file consistently check return values (e.g., `wc_SetSeed_Cb`, `wolfCrypt_SetPrivateKeyReadEnable_fips`).

```c
_lib.wolfCrypt_Init()
```

**Recommendation:** Check the return value and raise `WolfCryptError` on failure to match the existing convention in this file.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,122 @@
...
+            self._unregister()
+
+        def __del__(self) -> None:
+            self._unregister()
+
+        def callback(self, device_id: int, info: _ffi.CData) -> int:
+            print(f"{device_id=} algo = {ALGO_TYPE_NAME[info.algo_type]}")
```
JeremiahM37: **🟠 [Medium] Debug print() statements left in production callback dispatcher**

Multiple `print(...)` statements remain in the production callback dispatch path. These will fire on every crypto operation that goes through the callback (potentially in tight loops), polluting stdout, breaking quiet test runs, and being unconditional in user code. The codebase elsewhere (e.g., `random.py`, `hashes.py`) does not use print for runtime tracing.

```c
print(f"{device_id=} algo = {ALGO_TYPE_NAME[info.algo_type]}")
# ...
print(f"hash = {HASH_TYPE_NAME[info.hash.type]}")
print(f"{info.hash.data=} {info.hash.data_size=} {info.hash.digest=} {info.hash.u.sha256=}")
# ...
print("hash_update_callback")
# ...
print("hash_finalize_callback")
```

**Recommendation:** Remove all print() debug statements before merging. If diagnostic output is desired, switch to `logging` so users can opt in via log level.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Findings:** 14 total — 9 posted, 5 skipped
**7 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Importing wolfcrypt fails when CRYPTO_CB is disabled at build time** — `wolfcrypt/__init__.py:49 and wolfcrypt/cryptocb.py:33-58`
- [High] **Double unregister of device on context-manager exit + GC** — `wolfcrypt/cryptocb.py:73-77, 117-118`
- [Medium] **Return value of wolfCrypt_Init not checked** — `wolfcrypt/__init__.py:52`
- [Medium] **Debug print() statements left in production callback dispatcher** — `wolfcrypt/cryptocb.py:80, 84, 85, 107, 111`
- [Medium] **Hash digest dict KeyError on unmapped hash types** — `wolfcrypt/cryptocb.py:84, 90`
- [Medium] **Digest output buffer not size-validated** — `wolfcrypt/cryptocb.py:90`
- [Medium] **RNG output length not validated** — `wolfcrypt/cryptocb.py:96-97`
- [Medium] **No tests for hash callbacks despite PR claim** — `tests/test_cryptocb.py`
- [Low] **Free-floating string literal used as 'comment' in build_ffi.py is misleading** — `scripts/build_ffi.py:1376-1412`

### Skipped findings
- [Medium] ~~Anonymous-union cdef relies on '...' partial layout — may not match reality~~
- [Low] ~~ctx==NULL fallback would never be reached — wrong return code~~
- [Low] ~~Type annotations on overridable methods declare positional args with leading underscores~~
- [Low] ~~rng_callback signature passes raw cffi pointer with no type annotation~~
- [Info] ~~Stale exception import in __init__.py guard~~

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,122 @@
...
+            # _lib.wc_CryptoCb_InfoString(info)
+            try:
+                if info.algo_type == _lib.WC_ALGO_TYPE_HASH:
```
JeremiahM37, line 108: **🟠 [Medium] Hash digest dict KeyError on unmapped hash types**

`HASH_TYPE_NAME[info.hash.type]` and `DIGEST_SIZE[info.hash.type]` raise `KeyError` for any hash type not enumerated (e.g., `WC_HASH_TYPE_SHA224`, `WC_HASH_TYPE_MD5`, `WC_HASH_TYPE_SHAKE128/256`). A `KeyError` raised inside the C-invoked Python callback propagates out as a generic exception which cffi does not handle gracefully — it will be printed as an unhandled exception and the callback will return 0 (success) by default, leading to wolfCrypt continuing as if the device handled the operation while the buffer is uninitialized. The dispatcher should detect unsupported hash types and return `CRYPTOCB_UNAVAILABLE` so wolfCrypt falls back to its software implementation.

```c
if info.algo_type == _lib.WC_ALGO_TYPE_HASH:
    print(f"hash = {HASH_TYPE_NAME[info.hash.type]}")
    ...
    digest = self.hash_finalize_callback(device_id, info.hash.type)
    _ffi.buffer(info.hash.digest, DIGEST_SIZE[info.hash.type])[:] = digest
```

**Recommendation:** Use `.get()` with a default and return `CRYPTOCB_UNAVAILABLE` for unknown hash types. Also broaden the `except` clause in `callback()` to catch `(NotImplementedError, KeyError)` and return `CRYPTOCB_UNAVAILABLE` rather than letting the exception bubble through cffi.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,122 @@
...
+                    print(f"hash = {HASH_TYPE_NAME[info.hash.type]}")
+                    print(f"{info.hash.data=} {info.hash.data_size=} {info.hash.digest=} {info.hash.u.sha256=}")
+                    if info.hash.digest == _ffi.NULL:
+                        self.hash_update_callback(device_id, info.hash.type, bytes(_ffi.buffer(info.hash.data, info.hash.data_size)))
+                    else:
+                        digest = self.hash_finalize_callback(device_id, info.hash.type)
+                        _ffi.buffer(info.hash.digest, DIGEST_SIZE[info.hash.type])[:] = digest
```
JeremiahM37, line 125: **🟠 [Medium] Digest output buffer not size-validated**

`_ffi.buffer(info.hash.digest, DIGEST_SIZE[info.hash.type])[:] = digest` will raise `ValueError: buffer assignment size mismatch` if the user-supplied `hash_finalize_callback` returns a `bytes` object with the wrong length. As above, raising out of a cffi `def_extern` callback is not graceful. Validate the length and return `CRYPTOCB_UNAVAILABLE` (or raise a clearer wolfcrypt error) for mismatches.

```c
digest = self.hash_finalize_callback(device_id, info.hash.type)
_ffi.buffer(info.hash.digest, DIGEST_SIZE[info.hash.type])[:] = digest
```

**Recommendation:** Validate digest length before copying into the FFI buffer.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,122 @@
...
+                    return 0
+                if info.algo_type == _lib.WC_ALGO_TYPE_CIPHER:
+                    self.cipher_callback(device_id)
+                    return 0
+                if info.algo_type == _lib.WC_ALGO_TYPE_RNG:
+                    out = self.rng_callback(device_id, info.rng.rng, info.rng.sz)
+                    _ffi.buffer(info.rng.out, info.rng.sz)[:] = out
```
JeremiahM37, line 133: **🟠 [Medium] RNG output length not validated**

Same shape as the hash-digest issue: `_ffi.buffer(info.rng.out, info.rng.sz)[:] = out` will raise if the user-supplied `rng_callback` returns a `bytes` of the wrong length. The test happens to return correct-length data, so it never exercises the failure mode. A misbehaving subclass will produce a confusing exception originating in the cffi callback rather than a clean wolfcrypt error.

```c
out = self.rng_callback(device_id, info.rng.rng, info.rng.sz)
_ffi.buffer(info.rng.out, info.rng.sz)[:] = out
```

**Recommendation:** Validate length before copying.

scripts/build_ffi.py
```diff
@@ -1331,6 +1341,110 @@ def build_ffi(local_wolfssl, features):
...
+                    //struct {
+                    //    Des3*       des;
+                    //    byte*       out;
+                    //    const byte* in;
+                    //    word32      sz;
+                    //} des3;
+                    //void* ctx;
+                };
+            } cipher;
+        """
```
JeremiahM37: **🔵 [Low] Free-floating string literal used as 'comment' in build_ffi.py is misleading**

After `cdef += """ ... union { """` (line 1370-1374), there is a freestanding triple-quoted string (lines 1376-1412) describing the cipher portion of the union which is *not* assigned to anything — it is silently dropped. The next `cdef += """..."""` (line 1413) appends the hash struct, leaving the cipher member entirely absent from the cdef. This works because Python evaluates the standalone string and discards it, but at a glance it looks like the cipher chunk *is* added to `cdef`. A reader has to notice the missing `+=` to realize it is intentionally disabled.

```c
cdef += """
typedef struct {
    int algo_type; /* enum wc_AlgoType */
    union {
"""

"""
    struct {
        int type; /* enum wc_CipherType */
        ...
    } cipher;
"""
cdef += """
    struct {
        int type; /* enum wc_HashType */
        ...
```

**Recommendation:** Either delete the dead string outright or convert it to a real Python comment block with a note explaining why the cipher member is intentionally omitted (referring to the cffi/anonymous-union sizing issue from the PR description). Avoid relying on Python's silent discard of expression-statement string literals.

wolfcrypt/__init__.py
roberthdevries: Added a check.

wolfcrypt/cryptocb.py
roberthdevries: Converted to logging messages.

wolfcrypt/cryptocb.py
roberthdevries: Fixed.

wolfcrypt/cryptocb.py
roberthdevries: Added an explicit check.

wolfcrypt/cryptocb.py
roberthdevries: Added a check

scripts/build_ffi.py
roberthdevries: Converted to a comment block, explaining the issue with cffi.

wolfcrypt/cryptocb.py
roberthdevries: Most print statements have been removed as they were only used for testing/debugging/reverse engineering.

wolfcrypt/cryptocb.py
roberthdevries: The tracking of the registration is done in the wolfcrypt library itself. Unregistering a device twice is a noop.
`__del__` is not certain to be called at any time soon (unless a `gc.collect()` call is done explicitly) so if more fine-grained control is required to control the lifespan of the callback, a context manager is the typical solution in Python.

So the unregister in `__exit__` must not be removed.

JeremiahM37: wc_CryptoCb_UnRegisterDevice looks up by devId only — no ownership check (cryptocb.c:428). The double-unregister is only a no-op if the slot is still INVALID_DEVID at the second call. In the cross-instance case (cb1 exits → cb2 registers same devId → cb1 GC'd), cb1's del will find cb2's live slot and ClearDev it. A _registered flag in del avoids this without removing exit's unregister.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Final
+
+from typing_extensions import Self
+
+from wolfcrypt._ffi import ffi as _ffi
+from wolfcrypt._ffi import lib as _lib
+
+from wolfcrypt.exceptions import WolfCryptError
+
+ALGO_TYPE_NAME: Final = defaultdict(
```
JeremiahM37: Module-level dicts read `_lib.WC_ALGO_TYPE_*` / `_lib.WC_HASH_TYPE_*` at import, but those are only in the cdef when CRYPTO_CB is enabled — so `import wolfcrypt.cryptocb` fails with `AttributeError` before reaching the `if _lib.CRYPTO_CB_ENABLED:` guard. Move the dicts inside the guard, and make the `CryptoCallback` import in `wolfcrypt/__init__.py` conditional too.

dgarske:
Hi @roberthdevries nice work on this. See peer review feedback and also fix merge conflicts. Thanks

wolfcrypt/cryptocb.py
roberthdevries: Guard has been moved around almost all code in this module.
Also added guard for import in `__init__.py`.

wolfcrypt/cryptocb.py
roberthdevries: This scenario would still fail randomly because in that case the second register on the same device id would fail because it would still be registered. There are no real guarantees when the `__del__` method is called unless you manually call `gc.collect()`. This is not really a typical way of doing things as the recommended way to control resources like this is by using a context manager.
However, the way the crypto callback API is typically used is that you will register a device somewhere around application start-up and will leave it registered forever.
The dynamic case (registering/unregistering) is currently only used for testing.

So having a check as you propose is not very useful in my opinion.
Either:
1. Use a context manager if you want to use a register/unregister use case.
2. For a register once use case, you the `__del__` method *might* be called when the callback class instance goes out of scope.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,157 @@
...
+                    else:
+                        digest = self.hash_finalize_callback(device_id, info.hash.type)
+                        if len(digest) != DIGEST_SIZE[info.hash.type]:
+                            raise ValueError(
+                                f"Generated digest is expected to be {DIGEST_SIZE[info.hash.type]} bytes long, "
+                                f"but is {len(digest)} bytes long"
+                            )
+                        _ffi.buffer(info.hash.digest, DIGEST_SIZE[info.hash.type])[:] = digest
+                    return 0
+                if info.algo_type == _lib.WC_ALGO_TYPE_CIPHER:
```
dgarske: **🟠 [Medium] cipher_callback path returns success (0) without performing any cipher operation**

For `WC_ALGO_TYPE_CIPHER`, `callback()` calls `self.cipher_callback(device_id)` and then unconditionally `return 0`. The cipher fields of `wc_CryptoInfo` are intentionally not exposed in the cdef (commented out per the PR description), so a subclass overriding `cipher_callback` has no way to read inputs or write outputs, yet a no-op override would cause wolfCrypt to be told the cipher succeeded while no encryption/decryption was actually performed. Out of the box this is masked because the default `cipher_callback` raises `NotImplementedError` (caught -> `CRYPTOCB_UNAVAILABLE`), but exposing an overridable `cipher_callback` hook that returns success-without-effect is a footgun, especially for a crypto library where a silent no-op cipher is dangerous.

**Fix:** Until the cipher struct is actually exposed, either remove the `cipher_callback` hook or always return `CRYPTOCB_UNAVAILABLE` for cipher so wolfCrypt falls back to software, rather than returning 0/success for an operation that did nothing.

dgarske (commented):
## Skoll Code Review

**6 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Callback exceptions are converted to 0 (success) with unwritten output buffer** — `wolfcrypt/__init__.py:57-63, wolfcrypt/cryptocb.py:103-138`
- [Medium] **cipher_callback path returns success (0) without performing any cipher operation** — `wolfcrypt/cryptocb.py:125-127,149-150`
- [Medium] **Hash callbacks and error paths are untested; default-device-id test asserts nothing** — `tests/test_cryptocb.py:33-53`
- [Medium] **__del__ unregister not guarded against interpreter-shutdown errors** — `wolfcrypt/cryptocb.py:100-101,152-153`
- [Medium] **CRYPTO_CB cdef hash union depends on SHA/SHA256/SHA384/SHA512/SHA3 cdefs being present** — `scripts/build_ffi.py:1424-1440`
- [Low] **Inconsistent logging style in callback (f-string vs lazy %-formatting)** — `wolfcrypt/cryptocb.py:104,109`
- [Low] **wolfCrypt_Init() added with no corresponding cleanup** — `wolfcrypt/__init__.py:53-55`

tests/test_cryptocb.py
```diff
@@ -0,0 +1,53 @@
...
+from wolfcrypt.random import Random
+
+
+if not _lib.CRYPTO_CB_ENABLED:
+    pytest.skip("Crypto Callbacks not supported", allow_module_level=True)
+
+from wolfcrypt.cryptocb import CryptoCallback
+
+
+def test_default_device_id():
```
dgarske, line 63: **🟠 [Medium] Hash callbacks and error paths are untested; default-device-id test asserts nothing**

Only the RNG callback path is exercised. There is no coverage for `hash_update_callback` / `hash_finalize_callback` (the most complex branch in `callback()`, including the NULL-digest update vs finalize distinction and the digest-length validation), no coverage for the length-mismatch error branches (which is exactly where the success-on-exception bug above manifests), and no coverage for the `cipher` / unknown-algo fallback to `CRYPTOCB_UNAVAILABLE`. `test_default_device_id` only `print()`s the value and asserts nothing, so it cannot fail meaningfully.

**Fix:** Add tests for the hash update/finalize paths, the length-mismatch/error paths, and the unknown-algo fallback; make `test_default_device_id` assert on the return value.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,157 @@
...
+        def __enter__(self) -> Self:
+            return self
+
+        def __exit__(
+            self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
+        ) -> bool:
+            self._unregister()
+            return False
+
+        def __del__(self) -> None:
```
dgarske, line 102: **🟠 [Medium] __del__ unregister not guarded against interpreter-shutdown errors**

`__del__` calls `self._unregister()` which dereferences module-level `_lib`. During interpreter shutdown `_lib`/`wc_CryptoCb_UnRegisterDevice` may already be torn down, raising `AttributeError` (or worse) from a destructor, producing noisy 'Exception ignored in __del__' output. The existing `Random.__del__` in `wolfcrypt/random.py` explicitly guards this exact scenario with a try/except for shutdown. The new code should follow the same established pattern. (Also note `__exit__` followed by `__del__` calls `_unregister` twice; this is harmless since unregister is idempotent, but the shutdown guard is still warranted.)

**Fix:** Guard the destructor's unregister against shutdown-time teardown, matching the pattern already used in `Random.__del__`.

scripts/build_ffi.py
```diff
@@ -1335,6 +1345,117 @@ def build_ffi(local_wolfssl, features):
...
+        #             //    byte*       out;
+        #             //    const byte* in;
+        #             //    word32      sz;
+        #             //} des3;
+        #             //void* ctx;
+        #         };
+        #     } cipher;
+        # """
+
+        cdef += """
```
dgarske, line 1449: **🟠 [Medium] CRYPTO_CB cdef hash union depends on SHA/SHA256/SHA384/SHA512/SHA3 cdefs being present**

The `wc_CryptoInfo.hash.u` union added under `if features["CRYPTO_CB"]` references the C types `wc_Sha`, `wc_Sha256`, `wc_Sha384`, `wc_Sha512`, and `wc_Sha3*`. Those `typedef struct {...}` declarations are emitted only under their own feature guards (`if features["SHA"]`, `["SHA256"]`, etc., lines 845-908). If a build enables CRYPTO_CB while any of those hash features is disabled, the cdef will reference an undeclared type and ffibuilder.cdef()/compile will fail. The default build enables all of them so it works today, but the implicit coupling is fragile and undocumented.

**Fix:** Make the CRYPTO_CB hash-union cdef robust to disabled hash features, or document/assert the dependency so a non-default feature combination fails with a clear message instead of an opaque cdef error.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,157 @@
...
+            self._unregister()
+
+        def callback(self, device_id: int, info: _ffi.CData) -> int:
+            log.debug(f"{device_id=} algo = {ALGO_TYPE_NAME[info.algo_type]}")
```
dgarske: **🔵 [Low] Inconsistent logging style in callback (f-string vs lazy %-formatting)**

Line 104 uses an eagerly-interpolated f-string inside `log.debug(...)`, while line 109 uses lazy `%s` formatting. Mixing the two is inconsistent and the f-string form is what pylint's logging-fstring-interpolation flags (the file already disables several pylint checks at the top). Pick one style; lazy `%`-args avoid formatting cost when debug logging is disabled.

**Fix:** Use lazy `%`-style logging consistently across the callback.

wolfcrypt/__init__.py
```diff
@@ -46,8 +46,22 @@
 if top_level_py not in ["setup.py", "build_ffi.py"]:
     from wolfcrypt._ffi import ffi as _ffi
     from wolfcrypt._ffi import lib as _lib
+    if _lib.CRYPTO_CB_ENABLED:
+        from wolfcrypt.cryptocb import CryptoCallback
     from wolfcrypt.exceptions import WolfCryptApiError
 
+    ret = _lib.wolfCrypt_Init()
```
dgarske, line 65: **🔵 [Low] wolfCrypt_Init() added with no corresponding cleanup**

`wolfCrypt_Init()` is now called unconditionally on package import (a reasonable change, and required for the cryptocb registration to work), but there is no matching `wolfCrypt_Cleanup()` registered (e.g. via `atexit`). For a process-lifetime library this is generally acceptable, but it is worth a deliberate decision/comment so it is clear cleanup was intentionally omitted rather than forgotten.

**Fix:** Confirm the omission of `wolfCrypt_Cleanup()` is intentional; add a brief comment or an atexit cleanup if appropriate.

wolfcrypt/cryptocb.py
roberthdevries: Fixed.

roberthdevries (commented):
Not all issues are addressed. Will continue working on this after typing is merged so that this part can be typed as well.

dgarske:
@roberthdevries now that the other PR's have been merged please take another look at the remaining items. Thanks!

wolfcrypt/cryptocb.py
roberthdevries: Removed the `cipher_callback` definition. Can only be added back when the cffi code generation for the `wc_CryptoInfo` structure is made to work.

wolfcrypt/cryptocb.py
roberthdevries: Please explain how this is an issue. It is not clear to me how the C-extension gets unloaded before this module.
I am lacking in knowledge/experience about this specific failure mode.
Is this explained somewhere in CFFI documentation?

wolfcrypt/__init__.py
roberthdevries: Added a comment that it is not needed to call this function. This function initializes a global data structure that lives as long as the application and gets cleaned up together with the process.

tests/test_cryptocb.py
roberthdevries: Changed the `test_default_device_id` so that it checks that the expected value (according to the configuration used for the Python version).

Added test to check support for crypto callbacks for hash functions. For this some infrastructure had to be added to the `_Hash` class to support `device_id`s needed to call the callbacks.

scripts/build_ffi.py
roberthdevries: Changed to conditional inclusion in the union depending on enabled features.

dgarske (commented):
## Skoll Code Review

**Findings:** 8 total — 8 posted, 0 skipped
**8 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Import-time reference to _lib.INVALID_DEVID reintroduces issue #2659 regression** — `wolfcrypt/hashes.py:43`
- [High] **device_id silently ignored for Sha256/Sha384/Sha512 (callback never fires)** — `wolfcrypt/hashes.py:235-298`
- [High] **Uncaught callback exceptions report success (0) to wolfCrypt with an unfilled buffer** — `wolfcrypt/__init__.py:64-69`
- [Medium] **Sha3 constructor does not expose device_id despite _init supporting it** — `wolfcrypt/hashes.py:352-358`
- [Medium] **CryptoCallback.__del__ not guarded for interpreter shutdown; double-unregister and self-reference cycle** — `wolfcrypt/cryptocb.py:96-103`
- [Medium] **Missing test coverage for callback error paths and SHA-2 device_id** — `tests/test_cryptocb.py:1-87`
- [Low] **Hash update-vs-finalize discriminator drops input when both in and digest are set** — `wolfcrypt/cryptocb.py:112-126`
- [Low] **lib.pyi wc_HashInfo union field name does not match C member** — `wolfcrypt/_ffi/lib.pyi:592-601`

wolfcrypt/hashes.py
```diff
@@ -40,18 +40,18 @@ class _Hash(ABC):
     A **PEP 247: Cryptographic Hash Functions** compliant
     **Hash Function Interface**.
     """
-    def __init__(self, string: BytesOrStr | None = None) -> None:
+    def __init__(self, string: BytesOrStr | None = None, device_id: int = _lib.INVALID_DEVID) -> None:
```
dgarske, line 43: **🔴 [High] Import-time reference to _lib.INVALID_DEVID reintroduces issue #2659 regression**

The new default argument `device_id: int = _lib.INVALID_DEVID` is evaluated at module-import time (default parameter values are evaluated when the `def` in the `_Hash` class body executes). `INVALID_DEVID` is only emitted into the CFFI cdef when `ML_KEM` OR `ML_DSA` OR (newly) `CRYPTO_CB` is enabled (`scripts/build_ffi.py:1317`). In a wolfSSL build with none of those three (exactly the configuration the `build-no-pqc` CI job exercises: `--enable-cryptonly ... --disable-mlkem --disable-mldsa` and no `--enable-cryptocb`), `_lib.INVALID_DEVID` does not exist and `import wolfcrypt.hashes` raises `AttributeError`. This is the precise regression that issue #2659 fixed for `wolfcrypt/random.py` (which deliberately hardcodes `device_id: int = -2` instead of `_lib.INVALID_DEVID`). This PR re-introduces the exact failure in `hashes.py`, breaking hashing for every no-PQC/no-cryptocb build and the `build-no-pqc` regression job. Line 358 (`self._init(_lib.INVALID_DEVID)`) has the same dependency but only fails at Sha3 instantiation time; line 43 fails at import, which is worse.

**Fix:** Follow the established `random.py` pattern: use the literal `-2` for the default (and at hashes.py:358), or unconditionally declare `INVALID_DEVID` in the cdef so it is always available. Verify the `build-no-pqc` CI job passes.

wolfcrypt/hashes.py
```diff
@@ -232,7 +232,7 @@ def __del__(self) -> None:
                 self._delete(self._native_object)
 
         @override
-        def _init(self) -> int:
+        def _init(self, device_id: int) -> int:
```
dgarske, line 235: **🔴 [High] device_id silently ignored for Sha256/Sha384/Sha512 (callback never fires)**

`_Hash.__init__` now threads `device_id` into `_init(device_id)`, and the `Sha` class was updated to honor it via `wc_InitSha_ex(obj, NULL, device_id)`. However `Sha256._init`, `Sha384._init`, and `Sha512._init` still call the non-`_ex` initializers `wc_InitSha256(obj)` / `wc_InitSha384(obj)` / `wc_InitSha512(obj)`, which take no devId and default to `INVALID_DEVID`. As a result, `Sha256(device_id=11)` (and 384/512) accept the parameter but silently drop it, so the registered crypto callback is never invoked for those hash types. `cryptocb.py` advertises support for these types (they are present in `DIGEST_SIZE`/`HASH_TYPE_NAME`), so a user reasonably expects the callback to fire. Silently ignoring a security-relevant device-routing parameter is a correctness trap: the user believes offload/interception is active when it is not.

**Fix:** Add the `_ex` initializers to the cdef (`wc_InitSha256_ex`, `wc_InitSha384_ex`, `wc_InitSha512_ex`) and use them so device_id is honored. If SHA-2 device support is intentionally out of scope for this 'initial' PR, reject `device_id != INVALID_DEVID` for those classes (or raise) rather than silently ignoring it, and document the limitation.

wolfcrypt/__init__.py
```diff
@@ -33,21 +37,37 @@
...
 
+    # Only wolfCrypt_Init() is called here.
+    # Calling wolfCrypt_Cleanup() is not needed as the application exit() will clean up the entire process
+    # including any wolfcrypt data in any case.
+    ret = _lib.wolfCrypt_Init()
+    if ret < 0:
+        raise WolfCryptApiError("WolfCrypt_Init failed", ret)
+
+    if _lib.CRYPTO_CB_ENABLED:
+        @_ffi.def_extern()
```
dgarske: **🔴 [High] Uncaught callback exceptions report success (0) to wolfCrypt with an unfilled buffer**

`CryptoCallback.callback` only catches `NotImplementedError`; any other exception (e.g. the `ValueError` it deliberately raises when `len(out) != info.rng.sz` or the digest length is wrong) propagates out through the `@_ffi.def_extern()` function `py_wc_crypto_callback`. cffi's `def_extern` with no `error=`/`onerror=` argument prints the traceback to stderr and returns the default value 0 to the C caller. In the wolfCrypt crypto-callback convention 0 means 'handled successfully'. So a buggy user callback that returns the wrong number of bytes causes wolfCrypt to treat the operation as successful while the output buffer (`info.rng.out` / `info.hash.digest`) was never written. For RNG this yields uninitialized stack/heap memory being consumed as 'random' data. The length-check `raise ValueError(...)` is therefore not just ineffective, it is actively harmful because the error is swallowed into a success code.

**Fix:** Pass a negative `error=` (and/or `onerror=`) to `@_ffi.def_extern()` so an uncaught exception returns a hard failure to wolfCrypt, and/or wrap the body of `callback` in a broad `except Exception:` that logs and returns a negative error code instead of letting exceptions escape. Add a test that a wrong-length RNG/digest return does not report success.

wolfcrypt/hashes.py
```diff
@@ -355,7 +355,7 @@ def __init__(self, string: BytesOrStr | None = None, size: int = SHA3_384_DIGEST
             self.digest_size = size
             self._delete = self._SHA3_FREE.get(size)
             self._copy = self._SHA3_COPY.get(size)
-            ret = self._init()
+            ret = self._init(_lib.INVALID_DEVID)
```
dgarske: **🟠 [Medium] Sha3 constructor does not expose device_id despite _init supporting it**

`Sha3._init` was updated to forward `device_id` to `wc_InitSha3_*(obj, NULL, device_id)`, but `Sha3.__init__` has signature `(self, string=None, size=...)` with no `device_id` parameter and calls `self._init(_lib.INVALID_DEVID)` with a hardcoded value. So Sha3 can never actually route to a crypto callback, and its `_init` change is dead code from the public API's perspective. This is inconsistent with `Sha` (which does expose device_id) and with the DIGEST_SIZE table in cryptocb.py that lists SHA3 types.

**Fix:** Add a `device_id` parameter to `Sha3.__init__` (and `Sha3.new`) and pass it to `_init`, or explicitly document that Sha3 device support is not yet wired. Also note line 358's `_lib.INVALID_DEVID` shares the import/eval-time dependency called out above.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,153 @@
...
+            self.device_id = device_id
+            self.ctx = _ffi.new_handle(self)
+            ret = _lib.wc_CryptoCb_RegisterDevice(device_id, _lib.py_wc_crypto_callback, self.ctx)
+            if ret < 0:  # pragma: no cover
+                raise WolfCryptError(f"CryptoCb device registration error ({ret})")
+
+        def __enter__(self) -> Self:
+            return self
+
+        def __exit__(
```
dgarske, line 96: **🟠 [Medium] CryptoCallback.__del__ not guarded for interpreter shutdown; double-unregister and self-reference cycle**

Three related cleanup concerns: (1) `__del__` calls `_unregister()` -> `_lib.wc_CryptoCb_UnRegisterDevice(...)` with no protection against interpreter-shutdown teardown, whereas `random.py:__del__` deliberately wraps its native free in `try/except AttributeError` because `_lib` may be partially torn down at shutdown; this can emit noisy exceptions during finalization. (2) When used as a context manager, `__exit__` calls `_unregister()` and then `__del__` calls it again -> the device is unregistered twice (harmless in wolfSSL but indicates missing idempotency/guard). (3) `self.ctx = _ffi.new_handle(self)` plus `new_handle` keeping `self` alive forms a reference cycle (self -> ctx -> self); combined with a user-defined `__del__`, objects created without `with` are only reclaimed by the cyclic GC, so the device stays registered longer than the caller expects.

**Fix:** Guard `__del__` against shutdown like `random.py`, make `_unregister` idempotent (track a `_registered` flag), and consider clearing `self.ctx` on unregister to break the cycle for prompt cleanup.

tests/test_cryptocb.py
```diff
@@ -0,0 +1,87 @@
+# test_cryptocb.py
```
dgarske, line 1: **🟠 [Medium] Missing test coverage for callback error paths and SHA-2 device_id**

The new tests cover only the happy path for RNG (SHA-1 hash) and default_device_id. There is no coverage for: (a) a callback returning the wrong number of bytes (the `ValueError` guard in `callback` - important given the def_extern success-on-exception issue above); (b) an unsupported algo/hash type returning `CRYPTOCB_UNAVAILABLE`; (c) `Sha256`/`Sha384`/`Sha512` with `device_id` set (which would have surfaced that device_id is silently ignored for those classes); (d) the not-implemented default methods raising `NotImplementedError`. These gaps let the two SHA-2 and exception-handling bugs above pass CI.

**Fix:** Extend test_cryptocb.py to cover error paths, unsupported-algo fallthrough, and SHA-256/384/512 device routing.

wolfcrypt/cryptocb.py
```diff
@@ -0,0 +1,153 @@
...
+            self._unregister()
+
+        def callback(self, device_id: int, info: _lib.wc_CryptoInfo) -> int:
+            log.debug("device_id=%d algo_type = %s", device_id, ALGO_TYPE_NAME[info.algo_type])
+            try:
+                if info.algo_type == _lib.WC_ALGO_TYPE_HASH:
+                    if info.hash.type not in DIGEST_SIZE:
+                        return _lib.CRYPTOCB_UNAVAILABLE
+                    log.debug("hash_type = %s", HASH_TYPE_NAME[info.hash.type])
+                    if info.hash.digest == _ffi.NULL:
```
dgarske, line 112: **🔵 [Low] Hash update-vs-finalize discriminator drops input when both in and digest are set**

The callback distinguishes update from finalize purely via `info.hash.digest == _ffi.NULL`. wolfCrypt's one-shot hash paths (e.g. `wc_ShaHash`) can invoke the crypto-cb hash entry with BOTH a non-NULL input (`data`/`data_size`) and a non-NULL `digest` in a single call. In that case this code takes the `else` (finalize) branch and never delivers the input data to `hash_update_callback`, producing a digest over missing data. The wolfcrypt-py `_Hash` object happens to call update and final separately so the tested path is safe, but a one-shot caller would be mis-hashed.

**Fix:** Handle update and finalize independently within a single invocation (process input if present, then finalize if digest present), or confirm wolfCrypt never passes both for the supported hash types.

wolfcrypt/_ffi/lib.pyi
```diff
@@ -569,3 +588,36 @@ def wc_dilithium_sign_msg_with_seed(msg: bytes, msg_len: int, sig: BytePtr, sig_
 def wc_MlDsaKey_GetPrivLen(key: DilithiumKey, len: IntPtr) -> int: ...
 def wc_MlDsaKey_GetPubLen(key: DilithiumKey, len: IntPtr) -> int: ...
 def wc_MlDsaKey_GetSigLen(key: DilithiumKey, len: IntPtr) -> int: ...
+
+@dataclass
```
dgarske, line 592: **🔵 [Low] lib.pyi wc_HashInfo union field name does not match C member**

The type stub declares the SHA-1 union member as `sha: FFI.CData`, but the cdef in build_ffi.py names it `sha1` (`wc_Sha* sha1;`). The runtime code never accesses these union members so there is no functional impact, but the stub is inconsistent with the actual struct and would mislead anyone typing against `info.hash.sha1`.

**Fix:** Rename the stub field to `sha1` to match the cdef, or align the cdef member name with the stub.

dgarske (changes requested):
https://github.com/wolfSSL/wolfcrypt-py/pull/114#pullrequestreview-4190053551

roberthdevries:
Now that the `INVALID_DEVID` constant is always available, the need for the regression test for the no-pqc job becomes questionable.
Propose to remove it from the pipeline

wolfcrypt/hashes.py
roberthdevries: The second option is chosen. This removes the need for the magic value -2.

wolfcrypt/hashes.py
roberthdevries: Added `deviced_id` parameter to `Sha3.new` and `Sha3.__init__`

wolfcrypt/hashes.py
roberthdevries: Used the `_ex` variants of the hash init functions and passed the `device_id` to them.

wolfcrypt/__init__.py
roberthdevries: Added an `onerror` callback that checks the exception type for a `WolfCryptApiError` and then returns the error code of that exception, in all other cases it returns a generic -1 error code.
It will log an error message for debugging purposes with the original exception string.

Also added a test to check the correct handling of a `ValueError`.

wolfcrypt/_ffi/lib.pyi
roberthdevries: Renamed.

roberthdevries (commented):
Not all review comments addressed. More time needed.

wolfcrypt/cryptocb.py
roberthdevries: This issue baffles me as I have never seen this happen in any Python interpreter on Linux. Maybe this was an issue in Python 2? Or on an embedded Python? Please help me understand why this is an issue.
It also seems important enough to warrant a test that reproduces the issue to prevent regression.

wolfcrypt/cryptocb.py
roberthdevries: This is not how the callback mechanism works. I admit that this seems strange, but if you look at how this works internally in the wolfcrypt library you will see that this is exactly how you discriminate between update and finalize calls.

tests/test_cryptocb.py
roberthdevries: Added a Sha256,Sha384,Sha512 and a Sha3 test case. (good flow).
Added a test with an empty crypto callback instance for AesGcmStream that confirms that no behavior is changed.

PR #115, opened 2026-04-27
Convert python project configuration to pyproject.toml.

roberthdevries:
Modern python projects standardize on pyproject.toml to reduce the number of configuration files required for all tools.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,10 @@ classifiers = [
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Topic :: Security",
     "Topic :: Security :: Cryptography",
     "Topic :: Software Development"
@@ -91,7 +95,7 @@ indent-width = 4
 # Assume Python 3.10
 target-version = "py310"
 
-[lint]
+[tool.ruff.lint]
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
@@ -105,7 +109,7 @@ unfixable = []
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
-[format]
+[tool.ruff.format]
 # Like Black, use double quotes for strings.
 quote-style = "double"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,133 @@
+[project]
+name = "wolfcrypt"
+description = "Python module that encapsulates wolfSSL's crypto engine API."
+readme = "README.rst"
+requires-python = ">=3.10"
+authors = [
+    {name = "wolfSSL Inc.", email="info@wolfssl.com"},
+]
+license = "GPL-2.0-only OR LicenseRef-WolfSSL"
+license-files = ["LICENSING.rst"]
+keywords = ["wolfssl", "wolfcrypt", "security", "cryptography"]
+classifiers = [
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Topic :: Security",
+    "Topic :: Security :: Cryptography",
+    "Topic :: Software Development"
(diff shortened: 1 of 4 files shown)


JeremiahM37:
Has requires-python = ">=3.10", current setup.py advertises 3.6-3.9

pyproject.toml
```diff
@@ -0,0 +1,133 @@
...
+# This is currently disabled by default, but it is planned for this
+# to be opt-out in the future.
+docstring-code-format = false
+
+# Set the line length limit used when formatting code snippets in
+# docstrings.
+#
+# This only has an effect when the `docstring-code-format` setting is
+# enabled.
+docstring-code-line-length = "dynamic"
```
JeremiahM37: **🔴 [High] Ruff `[lint]` and `[format]` tables are at the top level instead of under `[tool.ruff]`**

Ruff reads its configuration from `[tool.ruff]`, `[tool.ruff.lint]`, and `[tool.ruff.format]` when configured via `pyproject.toml`. The diff places `[lint]` and `[format]` as top-level TOML tables, so ruff will silently ignore them — the `select`/`ignore` rules, `quote-style`, `indent-style`, etc. will not take effect. Only the `[tool.ruff]` section above (exclude/line-length/target-version) will be applied. This silently regresses lint/format behavior compared to a working `ruff.toml` or correctly-namespaced config.

```c
[lint]
# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
...
select = ["E4", "E7", "E9", "F", "B", "UP"]
ignore = ["UP031", "UP025", "UP032"]
...
[format]
quote-style = "double"
```

**Recommendation:** Rename the `[lint]` and `[format]` tables to `[tool.ruff.lint]` and `[tool.ruff.format]` so ruff actually picks up the configuration.

pyproject.toml
```diff
@@ -0,0 +1,133 @@
+[project]
+name = "wolfcrypt"
+description = "Python module that encapsulates wolfSSL's crypto engine API."
+readme = "README.rst"
+requires-python = ">=3.10"
```
JeremiahM37, line 5: **🟠 [Medium] `requires-python = ">=3.10"` silently drops Python 3.6–3.9 support**

The previous `setup.py` advertised support for Python 3.6, 3.7, 3.8, and 3.9 via classifiers (with no explicit `python_requires`). The new metadata bumps the floor to Python 3.10, which is a backward-incompatible change for downstream users still on 3.8/3.9 (still in widespread enterprise/distro use as of 2026). The PR description (`Modern python projects standardize on pyproject.toml…`) only mentions configuration consolidation, not a Python-version bump. If this is intentional it should be called out in the PR description and/or CHANGELOG; if not, restore the lower floor (e.g. `>=3.8`).

```c
requires-python = ">=3.10"
```

**Recommendation:** Confirm the Python 3.10 floor is intentional and document the deprecation in the changelog; otherwise lower the floor back to the previously-supported version.

JeremiahM37 (commented):
## Skoll Code Review

**Scan type:** `review`
**Findings:** 9 total — 2 posted, 7 skipped

### Posted findings
- [High] **Ruff `[lint]` and `[format]` tables are at the top level instead of under `[tool.ruff]`** — `pyproject.toml:94-133`
- [Medium] **`requires-python = ">=3.10"` silently drops Python 3.6–3.9 support** — `pyproject.toml:5`

### Skipped findings
- [High] ~~`cffi` missing from `[build-system].requires` — PEP 517 builds will fail~~
- [Medium] ~~`build-backend = setuptools.build_meta:__legacy__` contradicts the PR's goal~~
- [Medium] ~~`dynamic = ["version"]` declared without `[tool.setuptools.dynamic]` configuration~~
- [Low] ~~Classifiers list only Python 3.10 even though 3.11/3.12/3.13 are allowed~~
- [Medium] ~~`[build-system].requires = ["setuptools"]` not pinned to a version supporting SPDX `license` / `license-files`~~
- [Low] ~~Tox config in `pyproject.toml` may not be picked up because `package` is a string instead of a wheel-package env reference~~
- [Low] ~~`setup_requires` left in `setup.py` — dead code under PEP 517~~

pyproject.toml
roberthdevries: This is intentional as older Python version are end-of-life. Ref: https://devguide.python.org/versions/

A note will be added to the Release Notes.

I tested with Python versions 3.10-3.15 that the software passes the unit tests which means that the statement above is actually true.

There is already one other patches (not merged) that adds something to the ChangeLog. Adding something now will create a merge conflict, as is true for some other patches as well. It would help a lot if these patches were merged on short notice, to help preventing this in the near future.

dgarske:
Hi @roberthdevries please resolve merge conflicts. Thank you

dgarske (dismissed):

roberthdevries:
Conflicts are resolved

dgarske (approved):

PR #120, opened 2026-05-17
Use f string instead of percent formatting

roberthdevries:
`printf`-style string formatting has a number of quirks, and leads to less
readable code than using `str.format` calls or f-strings.
In general, prefer the newer `str.format` and f-strings constructs over
`printf`-style string formatting.


dgarske (approved):
Thank you @roberthdevries . Contributor agreement on file.

PR #125, opened 2026-05-21
Add typing: asn, exceptions, hashes, pwdbased, utils.

roberthdevries:


Trooper-X:
This also adds the `ruff` and `ty` dependencies.
Would be nice if this MR gets merged.

dgarske (commented):
## Skoll Code Review

**6 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **AES-SIV single-block associated-data length uses char count of original input, not encoded byte length** — `wolfcrypt/ciphers.py:397-400`
- [Medium] **sign_with_seed no longer accepts bytearray/memoryview seeds (regression)** — `wolfcrypt/ciphers.py:2513-2546`
- [Medium] **make_key_from_seed now silently UTF-8-encodes a str seed instead of rejecting it** — `wolfcrypt/ciphers.py:2360-2367`
- [Low] **ChaCha __init__ renamed size to _size, breaking the documented backward-compatible keyword** — `wolfcrypt/ciphers.py:544`
- [Low] **HKDF helpers annotate hash_cls as instance type instead of class type** — `wolfcrypt/hkdf.py:33,78,105`
- [Low] **Random no longer nulls native_object on init failure; __del__ frees an uninitialized RNG** — `wolfcrypt/random.py:37-52`
- [Low] **RsaPublic.__init__ made key a required positional argument** — `wolfcrypt/ciphers.py:771-774`

wolfcrypt/ciphers.py
```diff
@@ -362,9 +394,9 @@ def _prepare_associated_data(associated_data):
             if isinstance(associated_data, (str, bytes, bytearray, memoryview)):
                 # A single block is provided.
                 # Make sure we have bytes.
-                associated_data = t2b(associated_data)
+                associated_data_bytes = t2b(associated_data)
```
dgarske, line 408: **🔴 [High] AES-SIV single-block associated-data length uses char count of original input, not encoded byte length**

The refactor renamed the local variable but left the `assocSz` assignment pointing at the original input. `associated_data_bytes = t2b(associated_data)` creates a NEW bytes object, while line 400 still does `result[0].assocSz = len(associated_data)` against the ORIGINAL `associated_data`. Before this PR the code reassigned `associated_data = t2b(associated_data)`, so `len(associated_data)` was the encoded byte length and was correct. Now, when `associated_data` is a `str` (an explicitly documented single-block input type — see the docstring "a single str, bytes, bytearray, or memoryview") containing non-ASCII/multibyte UTF-8 characters, `len(associated_data)` is the character count, which is smaller than `len(associated_data_bytes)` (the UTF-8 byte count). `result[0].assoc` points at the full encoded byte buffer but `assocSz` is set too small, so `wc_AesSivEncrypt_ex`/`wc_AesSivDecrypt_ex` authenticate only a truncated prefix of the associated data. The library's own encrypt/decrypt round-trip will still pass (both sides truncate identically), masking the defect, but the tag will not match any spec-correct implementation and part of the AD goes unauthenticated. The list branch (line 412) correctly uses `len(block)` over `associated_data_bytes`, so only the single-block branch is affected.

**Fix:** Change line 400 to use `len(associated_data_bytes)` so the authenticated length matches the buffer actually pointed to. Add a test with a non-ASCII `str` single-block associated data (e.g. compare a str AD against the equivalent UTF-8 bytes AD) to lock in the fix.

wolfcrypt/ciphers.py
```diff
@@ -2474,13 +2510,7 @@ def sign_with_seed(self, message, seed, ctx=None):
...
             out_size[0] = in_size
 
-            try:
-                seed_view = memoryview(seed)
-            except TypeError as exception:
-                raise TypeError(
-                    "seed must support the buffer protocol, such as `bytes` or `bytearray`"
-                ) from exception
-            if len(seed_view) != ML_DSA_SIGNATURE_SEED_LENGTH:
+            if len(seed) != ML_DSA_SIGNATURE_SEED_LENGTH:
```
dgarske, line 2603: **🟠 [Medium] sign_with_seed no longer accepts bytearray/memoryview seeds (regression)**

The PR removed the `memoryview(seed)` + `_ffi.from_buffer(seed_view)` handling and now passes the raw `seed` object straight into `wc_dilithium_sign_msg_with_seed`/`wc_dilithium_sign_ctx_msg_with_seed`. The rest of this codebase consistently converts inputs to `bytes` (via `t2b`) or wraps them with `_ffi.from_buffer` before handing them to cffi, which strongly implies cffi accepts a plain `bytes` object for the `const byte*` argument but not a `bytearray`/`memoryview`. Under the old code a `bytearray` or `memoryview` seed of the correct length worked; now it will raise a raw cffi `TypeError`. This is a behavioral regression for buffer-protocol seed inputs, and the previous friendly error message ("seed must support the buffer protocol, such as bytes or bytearray") is also lost. The updated test only verifies that a `str` raises `TypeError`; there is no test for `bytearray`/`memoryview` seeds.

**Fix:** Restore buffer-protocol handling so bytes/bytearray/memoryview seeds all work while still rejecting str (e.g. `bytes(memoryview(seed))`). Add a parametrized test covering bytearray and memoryview seeds.

wolfcrypt/ciphers.py
```diff
@@ -2316,35 +2358,29 @@ def make_key_from_seed(cls, mldsa_type, seed):
             :type seed: bytes
             """
             mldsa_priv = cls(mldsa_type)
```
dgarske, line 2430: **🟠 [Medium] make_key_from_seed now silently UTF-8-encodes a str seed instead of rejecting it**

The PR replaced the `memoryview(seed)` buffer-protocol check with `seed_bytes = t2b(seed)`. `t2b` accepts `str` and encodes it as UTF-8, so a `str` seed (previously rejected with a `TypeError` because `memoryview(str)` fails) is now silently accepted and used as deterministic key material. This is a footgun for deterministic key generation and is inconsistent with the sibling `sign_with_seed`, which (per the updated test) now rejects `str` seeds with a `TypeError`. The two seed-consuming APIs should agree on whether `str` is a valid seed type.

**Fix:** Decide on one seed-type policy and apply it to both make_key_from_seed and sign_with_seed (recommend: accept bytes-like, reject str). Add a test asserting str seeds are rejected in make_key_from_seed.

wolfcrypt/ciphers.py
```diff
@@ -509,7 +541,7 @@ class ChaCha(_Cipher):
         _IV_nonce = b""
         _IV_counter = 0
 
-        def __init__(self, key="", size=32):  # pylint: disable=unused-argument
+        def __init__(self, key: BytesOrStr="", _size: int= 32) -> None:  # pylint: disable=unused-argument
```
dgarske: **🔵 [Low] ChaCha __init__ renamed size to _size, breaking the documented backward-compatible keyword**

The parameter `size` was renamed to `_size`, yet the in-code comment immediately below still states "size is kept for backwards compatibility". Renaming to `_size` breaks any caller that used the keyword form `ChaCha(key, size=32)` (now raises `TypeError: unexpected keyword argument 'size'`). Positional callers (e.g. the test at tests/test_ciphers.py:317 `ChaCha(vectors[ChaCha].key, 32)`) are unaffected, which is why the test suite still passes, but the public keyword is a contract change that contradicts the stated intent.

**Fix:** Keep the public keyword name `size` (the ANN/unused-arg lint can be silenced with the existing pylint disable / a leading-underscore rename only if the public keyword truly does not matter). If the keyword is genuinely deprecated, document that rather than silently renaming it.

wolfcrypt/random.py
```diff
@@ -34,11 +34,10 @@ class Random:
     """
 
     def __init__(self, nonce: __builtins__.bytes = b"", device_id: int = -2) -> None:
-        self.native_object: _lib.RNG | None = _ffi.new("WC_RNG *")
+        self.native_object: _lib.RNG = _ffi.new("WC_RNG *")
```
dgarske: **🔵 [Low] Random no longer nulls native_object on init failure; __del__ frees an uninitialized RNG**

The PR changed the attribute type to non-optional `_lib.RNG` and removed `self.native_object = None` from the `wc_InitRngNonce_ex` failure path. Because `__init__` assigns `native_object` before the init call, if init fails the object still has a truthy cdata, so `__del__` (which runs even when `__init__` raises) will call `wc_FreeRng` on a `WC_RNG` that failed to initialize. This is almost certainly benign (`_ffi.new` zero-fills and `wc_FreeRng` is null-safe on the DRBG), and the path is `# pragma: no cover`, but the removed assignment was a defensive guard. Note the failure path is the only place the old `| None` value could occur, so dropping the Optional type is otherwise fine.

**Fix:** Restore a null-out on the init-failure path (or allocate after a successful init) so `__del__` does not free an uninitialized RNG.

wolfcrypt/ciphers.py
```diff
@@ -731,7 +768,7 @@ def _get_mgf(self):
 
 
     class RsaPublic(_Rsa):
-        def __init__(self, key=None, hash_type=None, rng=None):
+        def __init__(self, key: BytesOrStr, hash_type: int | None = None, rng: Random | None = None) -> None:
```
dgarske, line 812: **🔵 [Low] RsaPublic.__init__ made key a required positional argument**

The signature changed from `def __init__(self, key=None, hash_type=None, rng=None)` to `def __init__(self, key: BytesOrStr, ...)`, dropping the `key=None` default and making `key` required, while the body still contains `if key is not None:`. In practice `RsaPublic()` already failed previously (with `key=None`, `len(key)` raised `TypeError`), so usable behavior is unchanged — but this is a public-API signature change and the residual `if key is not None` check is now effectively dead for the default path. `RsaPrivate` is unaffected since it does not call `RsaPublic.__init__`.

**Fix:** Confirm the API change is intentional; if so, remove the dead `if key is not None` branch, otherwise restore the default.

roberthdevries: The API change is intentional, as the method requires the key to be defined for `wc_RsaPublicKeyDecode` Passing a NULL pointer will return `BAD_FUNC_ARG` and `None` will raise a `TypeError`, so from that point of view it makes sense to not have a default `None` parameter.
The redundant check for `None` for the key is removed.

wolfcrypt/ciphers.py
roberthdevries: Fixed

wolfcrypt/ciphers.py
roberthdevries: Reverted the change.

wolfcrypt/ciphers.py
roberthdevries: After a bit of reverse engineering the set of accepted types are actually bytes, list/tuple of int (where each int should be in the range of values that a byte can have).
This is actually checked by the automatically generated bindings of cffi, and those bindings will perform the checks and will raise a `TypeError` if you pass a string for example.
This removes the need to check explicitly and aligns the accepted types of this method with the accepted types of the underlying call to `wc_dilithium_make_key_from_seed`

There is not yet an automated test. But manually testing shows that `str` is rejected.

wolfcrypt/random.py
roberthdevries: The original `None` value of `native_object` has been restored with a minor twist.
There is now an internal `_native_object` that is initialized to `None` during init.
If `_ffi.new` throws a `MemoryError` the `__del__` method now correctly detects that there is no native object to delete.
Also if `wc_InitRngNonce_ex` fails, the underlying native object is set to `None`.
But to prevent all users of `Random.native_object` to have to check for `None` which is a bit fragile, a property is introduced that checks for `None` and will raise a `WolfCryptError` exception.

wolfcrypt/ciphers.py
roberthdevries: After a bit of reverse engineering the set of accepted types are actually bytes, list/tuple of int (where each int should be in the range of values that a byte can have).
This is actually checked by the automatically generated bindings of cffi, and those bindings will perform the checks and will raise a TypeError if you pass a string for example.
This removes the need to check explicitly and aligns the accepted types of this method with the accepted types of the underlying call to `wc_dilithium_sign_ctx_msg_with_seed`

dgarske (commented):
## Skoll Code Review

### Posted findings
- [High] **New undeclared runtime dependency on typing_extensions** — `wolfcrypt/ciphers.py:28, wolfcrypt/hashes.py:28`
- [Medium] **Removed _ffi.from_buffer() drops bytearray/memoryview support for seed/rand inputs** — `wolfcrypt/ciphers.py:2383, 2548, 2559, 2067, 2110`
- [Medium] **_Cipher.new() dropped **kwargs, breaking PEP 272 extra keyword arguments** — `wolfcrypt/ciphers.py:187-199`
- [Medium] **HKDF functions annotate hash_cls as instance type instead of class type** — `wolfcrypt/hkdf.py:33, 78, 105`
- [Medium] **asn.py leaves function arguments unannotated while enabling ANN ruff rules** — `wolfcrypt/asn.py:81, 99`
- [Low] **test_mldsa now relies on cffi's low-level TypeError instead of an explicit guard** — `tests/test_mldsa.py:186`
- [Low] **RsaPublic.__init__ made key a required positional argument** — `wolfcrypt/ciphers.py:781`

wolfcrypt/ciphers.py
```diff
@@ -152,10 +157,35 @@ def __init__(self, key, mode, IV=None):
...
+    @abstractmethod
+    def _set_key(self, direction: int) -> int: ...
+
+    @abstractmethod
+    def _encrypt(self, destination: _ffi.CData, source: bytes) -> int: ...
+
+    @abstractmethod
+    def _decrypt(self, destination: _ffi.CData, source: bytes) -> int: ...
 
     @classmethod
```
dgarske, line 192: **🟠 [Medium] _Cipher.new() dropped **kwargs, breaking PEP 272 extra keyword arguments**

The signature changed from `def new(cls, key, mode, IV=None, **kwargs)` to `def new(cls, key, mode, IV=None) -> _Cipher`. The original `**kwargs` (kept with a `pylint: disable=W0613`) silently absorbed extra keyword arguments, which is part of the PEP 272 `new()` convention (callers may pass things like `counter` or `segment_size`). Removing it means any caller passing an extra keyword now gets a `TypeError`. This is a subtle, possibly unintended API narrowing.

**Fix:** Keep **kwargs for backward compatibility unless you intend to formally drop the PEP 272 extra-argument contract.

wolfcrypt/asn.py
```diff
@@ -75,7 +78,7 @@ def hash_oid_from_class(hash_cls):
         else:
             raise WolfCryptError(f"Unknown hash class {hash_cls.__name__}")
 
-    def make_signature(data, hash_cls, key=None):
+    def make_signature(data: bytes, hash_cls: type[_Hash], key = None) -> bytes:
```
dgarske: **🟠 [Medium] asn.py leaves function arguments unannotated while enabling ANN ruff rules**

This PR adds `"ANN"` to the ruff `select` list and only ignores it for `scripts/*.py` and `tests/*.py` (not `wolfcrypt/`). Yet `make_signature(data, hash_cls, key = None)` leaves `key` unannotated and `check_signature(signature, data, hash_cls, pub_key)` leaves `pub_key` unannotated. ruff's ANN001 (missing-type-function-argument) would flag both, so `ruff check` with the new config would report violations in a file the PR is explicitly typing. This is an internal inconsistency in the typing pass.

**Fix:** Complete the annotations in asn.py (and verify ruff ANN passes for wolfcrypt/) so the newly enabled rule does not fail lint.

tests/test_mldsa.py
```diff
@@ -183,7 +183,7 @@ def test_sign_with_seed(mldsa_type, rng):
 
         # test that the seed type is checked (should be bytes-like, not string)
         with pytest.raises(TypeError):
-            _ = mldsa_priv.sign_with_seed(message, "")
+            _ = mldsa_priv.sign_with_seed(message, " " * ML_DSA_SIGNATURE_SEED_LENGTH)
```
dgarske: **🔵 [Low] test_mldsa now relies on cffi's low-level TypeError instead of an explicit guard**

The removed code in `sign_with_seed`/`make_key_from_seed` previously raised a clear, library-owned TypeError ('seed must support the buffer protocol...') when given a non-buffer seed. With that guard gone, the test was changed from `sign_with_seed(message, "")` to `sign_with_seed(message, " " * ML_DSA_SIGNATURE_SEED_LENGTH)` so that the length check passes and the str then reaches CFFI, which raises TypeError. The test still passes, but it now asserts on CFFI's internal type rejection ('must be a bytes or list or tuple, not str') rather than a wolfcrypt-controlled error. This couples the test to CFFI internals and produces a less user-friendly error for callers.

**Fix:** Optional: re-add an explicit type guard for clearer errors; otherwise acceptable as-is.

wolfcrypt/ciphers.py
```diff
@@ -731,18 +778,16 @@ def _get_mgf(self):
 
 
     class RsaPublic(_Rsa):
-        def __init__(self, key=None, hash_type=None, rng=None):
+        def __init__(self, key: BytesOrStr, hash_type: int | None = None, rng: Random | None = None) -> None:
```
dgarske, line 812: **🔵 [Low] RsaPublic.__init__ made key a required positional argument**

The signature changed from `__init__(self, key=None, hash_type=None, rng=None)` to `__init__(self, key: BytesOrStr, hash_type=None, rng=None)`, removing the default for `key`. `RsaPublic()` / `RsaPublic(key=None)` now raises TypeError/`t2b(None)` TypeError instead of the previous (already-broken) path that crashed at `len(key)`. All in-repo callers (and `RsaPrivate`, which uses its own __init__) pass a key, so impact is minimal, but it is a public signature change worth noting in case external code instantiated RsaPublic without a key.

**Fix:** Low priority; document the signature change. No functional regression for existing callers.

tests/test_mldsa.py
roberthdevries: Added an explicit check for the allowed types. However if passing a list/tuple with integers exceeding the value 255 the cffi internal check will still trigger.

wolfcrypt/ciphers.py
roberthdevries: Re-added `**kwargs` with type `int` as all (optional) arguments in PEP 272 have that type.

wolfcrypt/ciphers.py
roberthdevries: Added the signature change to the release notes.

wolfcrypt/asn.py
roberthdevries: Added the missing annotations.

wolfcrypt/ciphers.py
```diff
@@ -2004,7 +2068,7 @@ def encapsulate_with_random(self, rand):
             ct = _ffi.new(f"unsigned char[{ct_size}]")
             ss = _ffi.new(f"unsigned char[{ss_size}]")
             ret = _lib.wc_KyberKey_EncapsulateWithRandom(
```
dgarske, line 2107: **🔵 [Low] ML-KEM *_with_random helpers no longer accept bytearray/memoryview for rand**

These calls dropped `_ffi.from_buffer(rand)` and now pass `rand` directly to `wc_KyberKey_EncapsulateWithRandom` / `wc_KyberKey_MakeKeyWithRandom` (`const unsigned char*`). Unlike `decode_key`/`decapsulate`, `rand` is not run through `t2b`, so a `bytes` value still works but a `bytearray` or `memoryview` (which `from_buffer` previously handled) now raises a cffi `TypeError`. The annotation was narrowed to `rand: bytes`, which documents the intent, but it is still a silent runtime behavior change for callers passing other buffer types. Tests only pass `bytes` (`unhexlify(...)`), so the change is untested.

**Fix:** Either normalize `rand` with `t2b()` (matching the other ML-KEM methods) to preserve buffer-protocol support, or explicitly document that only `bytes` is accepted now.

wolfcrypt/hkdf.py
```diff
@@ -28,8 +28,9 @@
 
 
 if _lib.HKDF_ENABLED:
```
dgarske, line 38: **🔵 [Low] hkdf.py forces a runtime import of _Hmac for a type annotation (no from __future__ import annotations)**

Unlike the other modified modules (asn.py, ciphers.py, hashes.py, pwdbased.py), hkdf.py does not add `from __future__ import annotations`. As a result the new annotation `def HKDF(hash_cls: type[_Hmac], ...)` is evaluated at import time, which is why a real runtime import `from wolfcrypt.hashes import _Hmac` was added inside `if _lib.HKDF_ENABLED:`. `_Hmac` is only defined when `_lib.HMAC_ENABLED`. In a build where HKDF is enabled but HMAC is not, `import wolfcrypt.hkdf` would now raise ImportError where it previously succeeded. wolfSSL's HKDF structurally depends on HMAC so this is unlikely in practice, but the new import coupling is avoidable.

**Fix:** Add `from __future__ import annotations` to hkdf.py and move the `_Hmac` import under `TYPE_CHECKING` (or guard it on `_lib.HMAC_ENABLED`) so the annotation does not create a hard runtime dependency.

wolfcrypt/ciphers.py
```diff
@@ -2306,45 +2369,38 @@ def make_key(cls, mldsa_type, rng=None):
             return mldsa_priv
 
         @classmethod
```
dgarske, line 2420: **🟠 [Medium] ML-DSA seed handling: bytearray/memoryview now rejected, and the two seed methods validate inconsistently**

Before this PR both methods accepted any buffer-protocol object (bytes, bytearray, memoryview) via `memoryview(seed)` followed by `_ffi.from_buffer(seed_view)`. The PR removes that normalization and the two methods now diverge:

- `make_key_from_seed` only does `len(seed)` and passes `seed` straight to `wc_dilithium_make_key_from_seed(...)` (a `const byte*` argument). CFFI accepts `bytes`, `list` and `tuple` for that argument but raises `TypeError` for `bytearray` and `memoryview` (verified empirically against `_cffi_backend`). So a bytearray/memoryview seed that worked before now fails with an opaque cffi error.
- `sign_with_seed` adds an explicit `if not isinstance(seed, (list, tuple, bytes)): raise TypeError(...)`, which also rejects bytearray/memoryview but with a different (clearer) error.

Net effect: the accepted seed set changed from {bytes, bytearray, memoryview} to {bytes, list, tuple}. `bytes` (the common path, and the only one exercised by tests) still works, but bytearray/memoryview are a silent regression, and the two methods now raise different error types for the same bad input. No test covers list/tuple seeds or bytearray/memoryview seeds, so CI will not catch the change.

**Fix:** Apply the same validation/normalization in both methods. If the input-type contract is intentionally being narrowed, note it in ChangeLog.rst and add tests covering list/tuple (and rejected bytearray) seeds so the behavior is pinned.

---
> **Note:** Referenced line (`wolfcrypt/ciphers.py:2371-2392 (make_key_from_seed), 2516-2540 (sign_with_seed)`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **ML-DSA seed handling: bytearray/memoryview now rejected, and the two seed methods validate inconsistently** — `wolfcrypt/ciphers.py:2371-2392 (make_key_from_seed), 2516-2540 (sign_with_seed)`
- [Low] **ML-KEM *_with_random helpers no longer accept bytearray/memoryview for rand** — `wolfcrypt/ciphers.py:2059-2077 (encapsulate_with_random), 2103-2119 (make_key_with_random)`
- [Low] **hkdf.py forces a runtime import of _Hmac for a type annotation (no from __future__ import annotations)** — `wolfcrypt/hkdf.py:30-33`

wolfcrypt/ciphers.py
roberthdevries: The intent of the original author appears to be that this parameter is meant to be `bytes` in all cases according to the unchanged doc-string of these methods. That the original implementation uses `_ffi.from_buffer` to convert bytes to a native cffi `char *` could be because in other locations the same convention was used and the author was unaware that Python `bytes` objects are automatically converted by the CFFI glue code.
So the explicit documentation is already that it accepts bytes, and now it is more reliant on that that is true.

wolfcrypt/ciphers.py
roberthdevries: The explicit type check is added in this method as well. The original doc-string states that the type of `seed` is bytes (in both cases). This has been expanded to allow other types that are accepted by the CFFI binding.
So no originally intended generality has been lost, only accidentally introduced generality.

wolfcrypt/hkdf.py
roberthdevries: Added the `from __future__ import annotations` and the gard (`if _lib.HMAC_ENABLED) to hkdf.py

wolfcrypt/ciphers.py
```diff
@@ -2306,45 +2369,42 @@ def make_key(cls, mldsa_type, rng=None):
...
             mldsa_priv = cls(mldsa_type)
-            try:
-                seed_view = memoryview(seed)
-            except TypeError as exception:
-                raise TypeError(
-                    "seed must support the buffer protocol, such as `bytes` or `bytearray`"
-                ) from exception
-            if len(seed_view) != cls.ML_DSA_KEYGEN_SEED_LENGTH:
+
+            if not isinstance(seed, (list, tuple, bytes)):
```
dgarske: **🟠 [Medium] ML-DSA seed validation now rejects bytearray/memoryview (regression)**

The seed validation was rewritten from a buffer-protocol check to an `isinstance(seed, (list, tuple, bytes))` check. The previous code did `memoryview(seed)` and explicitly documented bytearray support (its error message read "seed must support the buffer protocol, such as `bytes` or `bytearray`"). The new check accepts only `list`, `tuple`, and `bytes`, so a caller passing a `bytearray` or `memoryview` seed - both previously valid - now gets `TypeError: seed must be bytes or list/tuple`. For a crypto API this is a real backward-incompatible narrowing, and `bytearray` is the natural choice for zeroizable secret seeds. This affects both `MlDsaPrivate.make_key_from_seed` and `MlDsaPrivate.sign_with_seed`.

**Fix:** Restore buffer-protocol support by normalizing the seed with `bytes(seed)` (which also handles list/tuple) before the length check and C call, instead of an isinstance allow-list that drops bytearray/memoryview.

wolfcrypt/ciphers.py
```diff
@@ -2306,45 +2369,42 @@ def make_key(cls, mldsa_type, rng=None):
             return mldsa_priv
 
         @classmethod
-        def make_key_from_seed(cls, mldsa_type, seed):
+        def make_key_from_seed(cls, mldsa_type: MlDsaType, seed: bytes | list[int] | tuple[int]) -> MlDsaPrivate:
```
dgarske: **🟠 [Medium] Advertised list/tuple seed support is untested and likely fails at the cffi boundary**

The docstrings and the new `lib.pyi` stub now advertise `seed: bytes | list[int] | tuple[int]`, and the validation lets `list`/`tuple` pass through to `_lib.wc_dilithium_make_key_from_seed(native_object, seed)` / `wc_dilithium_sign_*_with_seed(...)`. The cdef in `scripts/build_ffi.py` declares the seed parameter as `const byte* seed`. Standard cffi does not auto-convert a Python `list`/`tuple` to a pointer argument (only `bytes` is accepted for `char*`/`byte*`; lists require `ffi.new("byte[]", seed)`), so passing a list would likely raise an opaque cffi `TypeError` at the call rather than working as documented. The test suite only exercises `bytes` seeds (`rng.bytes(...)`, `bytes(N)`) and the `""` rejection case, so this newly-advertised path is entirely uncovered.

**Fix:** Either normalize the seed to `bytes` before the C call (fixes both this and the bytearray regression) or drop the list/tuple claim. Add a unit test that passes a `list[int]` seed to `make_key_from_seed`/`sign_with_seed` so the advertised type is actually verified.

setup.py
```diff
@@ -63,5 +63,5 @@
     install_requires=["cffi>=1.0.0,<2"],
```
dgarske: **🔵 [Low] setup.py install_requires not synced with new typing-extensions runtime dependency**

This PR adds `typing-extensions` to `[project].dependencies` in `pyproject.toml` and now imports it unconditionally at module load time (`from typing_extensions import override` in both `wolfcrypt/ciphers.py` and `wolfcrypt/hashes.py`). `override` only exists in the stdlib `typing` from Python 3.12+, and the project supports 3.10/3.11, so `typing-extensions` is a hard runtime dependency. `setup.py` still lists `install_requires=["cffi>=1.0.0,<2"]` only. The project specifies dependencies in both `pyproject.toml` and `setup.py`; whichever path is used for install metadata, leaving the two out of sync is confusing and risks an `ImportError` for any install path that honors `setup.py`'s `install_requires`.

**Fix:** Add `typing-extensions` to `setup.py`'s `install_requires` to match `pyproject.toml`, or remove the duplicate dependency list from `setup.py` so `pyproject.toml` is the single source of truth.

wolfcrypt/__init__.py
```diff
@@ -17,6 +17,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA
+from typing import cast
+
+from _cffi_backend import Lib
```
dgarske: **🔵 [Low] Hard runtime import of private cffi symbol _cffi_backend.Lib only to satisfy a cast**

`wolfcrypt/__init__.py` now does `from _cffi_backend import Lib` at module top and uses it as `_ffi.addressof(cast(Lib, _lib), "wc_GenerateSeed")`. Because `typing.cast(Lib, _lib)` evaluates its first argument at runtime, `Lib` must be importable whenever `wolfcrypt` is imported. This couples package import to a private cffi implementation detail (`_cffi_backend.Lib`) purely to silence a type checker; the cast has no runtime effect on behavior. If a future cffi release renames/removes that symbol, importing the whole package breaks for zero functional benefit.

**Fix:** Guard the `Lib` import under `TYPE_CHECKING` and use a `# type: ignore` (or a string-annotated local) instead of a runtime `cast`, so package import does not depend on a private cffi symbol.

dgarske (commented):
## Skoll Code Review

**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [Medium] **ML-DSA seed validation now rejects bytearray/memoryview (regression)** — `wolfcrypt/ciphers.py:2383,2537`
- [Medium] **Advertised list/tuple seed support is untested and likely fails at the cffi boundary** — `wolfcrypt/ciphers.py:2372,2391`
- [Low] **setup.py install_requires not synced with new typing-extensions runtime dependency** — `setup.py:62-63`
- [Low] **Hard runtime import of private cffi symbol _cffi_backend.Lib only to satisfy a cast** — `wolfcrypt/__init__.py:20-22,56`

dgarske (changes requested):
See https://github.com/wolfSSL/wolfcrypt-py/pull/125#pullrequestreview-4517797972
If you disagree just make note. We are getting close on this and thank you for your efforts

setup.py
roberthdevries: Removed the `install_requires` from setup.py to make pyproject.toml the single source of truth.

wolfcrypt/ciphers.py
roberthdevries: `seed` is now forced to be compatible to `bytes` before passing it to `wc_dilithium_make_key_from_seed`

wolfcrypt/ciphers.py
roberthdevries: `seed` is now forced to be compatible with `bytes` before passing it on to `wc_dilithium_make_key_from_seed`

wolfcrypt/__init__.py
roberthdevries: Added a `ty: ignore` to remove the need for casting.

dgarske (dismissed):

pyproject.toml
```diff
@@ -24,6 +24,7 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "cffi>=1.0.0,<2",
+    "typing-extensions",
```
dgarske: **🟠 [Medium] typing-extensions dependency has no minimum version (override requires =4.4.0)**

The PR adds `from typing_extensions import override` as an unconditional runtime import in `wolfcrypt/ciphers.py` and `wolfcrypt/hashes.py`. `override` was only added to `typing_extensions` in 4.4.0 (and `typing.override` does not exist until Python 3.12, while this package now supports 3.10/3.11 per `requires-python = ">=3.10"`). The new dependency is declared as just `"typing-extensions"` with no version floor. On an environment that already has an older `typing_extensions` installed, pip will treat the unconstrained requirement as already satisfied and not upgrade it, so `from typing_extensions import override` raises `ImportError` and the *entire* `wolfcrypt` package fails to import.

**Fix:** Pin a minimum version (`typing-extensions>=4.4.0`) so `override` is guaranteed to be importable on all supported Python versions.

wolfcrypt/ciphers.py
```diff
@@ -2316,35 +2379,31 @@ def make_key_from_seed(cls, mldsa_type, seed):
             :type seed: bytes
             """
             mldsa_priv = cls(mldsa_type)
```
dgarske, line 2430: **🔵 [Low] bytes(seed) silently accepts an int, dropping the friendly type check for ML-DSA seeds**

The previous code wrapped the seed in `memoryview(seed)` inside a try/except that raised a clear `TypeError("seed must support the buffer protocol, such as `bytes` or `bytearray`")` for non-buffer inputs. The new code replaces that with `seed = bytes(seed)`. For `str`/`None` this still raises `TypeError` (acceptable), but for an `int` argument `bytes(n)` silently produces `n` zero bytes. If a caller mistakenly passes the integer seed length (e.g. `make_key_from_seed(t, 32)`), the length check `len(seed) != ML_DSA_KEYGEN_SEED_LENGTH` passes and an all-zero deterministic seed is used to generate a key/signature with no error -- a subtle footgun for a cryptographic seed input. The original explicit, descriptive error message is also lost.

**Fix:** Reject `int` (and other non-buffer scalar) inputs explicitly, or restore a buffer-protocol type check, so a misused integer cannot be silently coerced into an all-zero seed.

wolfcrypt/types.py
```diff
@@ -0,0 +1,35 @@
+# types.py
```
dgarske: **⚪ [Info] New module wolfcrypt/types.py shadows the stdlib types module name**

The PR introduces `wolfcrypt/types.py`. It is consumed only via package-relative imports (`from .types import ...`), and Python 3 absolute imports prevent it from shadowing the stdlib `types` module elsewhere, so there is no runtime bug. However, naming a module `types` is a well-known footgun for maintainability (any future `import types` inside the package can confuse readers/tools).

**Fix:** Consider renaming the module to avoid colliding with the stdlib `types` name; optional, no functional impact.

wolfcrypt/__init__.py
```diff
@@ -50,7 +50,7 @@
 
     if hasattr(_lib, 'WC_RNG_SEED_CB_ENABLED'):
         if _lib.WC_RNG_SEED_CB_ENABLED:
-            ret = _lib.wc_SetSeed_Cb(_ffi.addressof(_lib, "wc_GenerateSeed"))
+            ret = _lib.wc_SetSeed_Cb(_ffi.addressof(_lib, "wc_GenerateSeed"))  # ty:ignore[no-matching-overload]
```
dgarske: **⚪ [Info] Inconsistent # ty:ignore comment lacks the space used everywhere else**

Every other suppression added in this PR uses the form `# ty: ignore[...]` (with a space after the colon), but this one is written `# ty:ignore[no-matching-overload]` (no space). If the `ty` checker's comment parser requires the canonical `# ty: ignore` spelling, this suppression will not take effect and the overload diagnostic will still surface. Cosmetic/tooling only -- no runtime impact.

**Fix:** Normalize to `# ty: ignore[no-matching-overload]` to match the rest of the codebase and ensure the suppression is recognized.

dgarske (commented):
## Skoll Code Review

**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [Medium] **typing-extensions dependency has no minimum version (override requires =4.4.0)** — `pyproject.toml:27`
- [Low] **bytes(seed) silently accepts an int, dropping the friendly type check for ML-DSA seeds** — `wolfcrypt/ciphers.py:2383,2536`
- [Info] **New module wolfcrypt/types.py shadows the stdlib types module name** — `wolfcrypt/types.py:1`
- [Info] **Inconsistent # ty:ignore comment lacks the space used everywhere else** — `wolfcrypt/__init__.py:53`

wolfcrypt/__init__.py
roberthdevries: Fixed

pyproject.toml
roberthdevries: Fixed

wolfcrypt/types.py
roberthdevries: Renamed to wc_types.py

wolfcrypt/ciphers.py
roberthdevries: Restored the buffer protocol check using `memoryview` without using the result of memoryview.
Instead the conversion to bytes will take care of passing the right datatype to the cffi wrapped function.

dgarske (approved):

tests/test_ciphers.py
```diff
@@ -320,7 +319,7 @@ def chacha_obj(vectors):
         return r
 
     @pytest.fixture
```
dgarske: **🟠 [Medium] ChaCha20 encrypt/decrypt round-trip test never executes (decorated as a fixture)**

The diff edits the signature of `test_chacha_enc_dec` from `(chacha_obj)` to `(chacha_obj, vectors)` (adding the `vectors` fixture that its body references at line 325). However, the function is decorated with `@pytest.fixture`, so pytest registers it as a *fixture* named `test_chacha_enc_dec` and never collects it as a test. As a result the ChaCha20 encrypt->set_iv->decrypt round-trip assertion (`assert plaintext == dec`) is dead code and provides zero coverage. The signature change indicates awareness that the body needs `vectors`, but the underlying problem is the `@pytest.fixture` decorator on a `test_`-prefixed function. (The `@pytest.fixture` line is pre-existing, but the PR touches this exact function.)

**Fix:** Remove the `@pytest.fixture` decorator so the function is collected and run as an actual test. Since the diff already modified this line to fix the missing `vectors` parameter, this is the natural moment to also make the test execute.

wolfcrypt/ciphers.py
```diff
@@ -730,19 +779,17 @@ def _get_mgf(self):
 
 
 
-    class RsaPublic(_Rsa):
-        def __init__(self, key=None, hash_type=None, rng=None):
+    class RsaPublic(_Rsa, SupportsRsaVerify):
```
dgarske, line 811: **⚪ [Info] RsaPublic constructor key parameter is now mandatory (breaking API change)**

`RsaPublic.__init__` changed from `def __init__(self, key=None, ...)` (with an internal `if key is not None:` guard) to `def __init__(self, key: BytesOrStr, ...)` where `key` is required and unconditionally passed to `t2b(key)`/`wc_RsaPublicKeyDecode`. Any caller that previously did `RsaPublic()` now raises `TypeError`. This is an intentional, documented change (ChangeLog: "The RsaPublic key parameter is now mandatory..."), and `RsaPrivate.__init__` still accepts `key=None` and uses `_Rsa.__init__` directly, so internal keygen paths (`RsaPrivate.make_key`) are unaffected. Noting for API-compatibility awareness only.

**Fix:** No change required; the breaking change is intentional and documented in ChangeLog.rst. Ensure release notes make the new mandatory-key requirement prominent for downstream users.

dgarske (commented):
## Skoll Code Review

**Findings:** 2 total — 2 posted, 0 skipped

### Posted findings
- [Medium] **ChaCha20 encrypt/decrypt round-trip test never executes (decorated as a fixture)** — `tests/test_ciphers.py:321-327`
- [Info] **RsaPublic constructor key parameter is now mandatory (breaking API change)** — `wolfcrypt/ciphers.py:782-799`

dgarske (changes requested):
Please also resolve merge conflicts. 

https://github.com/wolfSSL/wolfcrypt-py/pull/125#pullrequestreview-4620864282

wolfcrypt/ciphers.py
roberthdevries: This is already in the release notes.

tests/test_ciphers.py
roberthdevries: This was not part of my MR, but fixed nonetheless.

wolfcrypt/ciphers.py
```diff
@@ -2028,7 +2094,7 @@ def encapsulate_with_random(self, rand):
             ct = _ffi.new(f"unsigned char[{ct_size}]")
             ss = _ffi.new(f"unsigned char[{ss_size}]")
             ret = _lib.wc_KyberKey_EncapsulateWithRandom(
```
dgarske, line 2107: **🔵 [Low] MlKem *_with_random drops buffer-protocol support for rand**

The PR replaces `_ffi.from_buffer(rand)` with passing `rand` straight into the CFFI call (`wc_KyberKey_EncapsulateWithRandom(self.native_object, ct, ss, rand, len(rand))` and the analogous `wc_KyberKey_MakeKeyWithRandom(..., rand, len(rand))`). For every other converted call site the argument is guaranteed to be `bytes` first (via `t2b(...)` for `decode_key`/`decapsulate`, or `seed = bytes(seed)` in the ML-DSA `make_key_from_seed`/`sign_with_seed` paths). `rand` here is the one exception: it is passed through unconverted. `_ffi.from_buffer` previously accepted any buffer-protocol object (`bytearray`, `memoryview`); CFFI's implicit conversion for a `const unsigned char *` argument only accepts `bytes`, so a caller that previously passed a `bytearray`/`memoryview` may now hit a `TypeError`. This is a silent, undocumented narrowing of a public API and is inconsistent with the ML-DSA seed handling in the same PR.

**Fix:** Normalize `rand` with `bytes(rand)` (mirroring the ML-DSA `bytes(seed)` conversion) before the CFFI call in both `encapsulate_with_random` and `make_key_with_random`, so `bytearray`/`memoryview` inputs keep working. The existing ML-KEM tests only pass `bytes` (`h2b(...)`), so this narrowing is not covered.

setup.py
```diff
@@ -63,5 +63,5 @@
     install_requires=["cffi>=1.17"],
```
dgarske: **🔵 [Low] setup.py dependencies out of sync with pyproject (typing-extensions)**

This PR makes `typing_extensions` a hard runtime import (`from typing_extensions import override` at module load in both `wolfcrypt/ciphers.py` and `wolfcrypt/hashes.py`) and correctly adds `typing-extensions>=4.4.0` to `[project].dependencies` in pyproject.toml. However, setup.py still declares `setup_requires=["cffi>=1.17"]` / `install_requires=["cffi>=1.17"]` with no mention of typing-extensions. Because the build backend is `setuptools.build_meta:__legacy__` and `[project].dependencies` is present, the pyproject metadata is authoritative for PEP 517 wheel builds, so pip installs will still pull typing-extensions - i.e. this is not a functional break. It is a consistency issue: the two dependency lists now disagree, and anyone using the deprecated `python setup.py install` path or reading setup.py as the source of truth would miss the new required dependency.

**Fix:** Add `typing-extensions>=4.4.0` to setup.py's install_requires to keep it in sync with pyproject.toml, or drop the redundant install_requires entirely and rely solely on the [project] table.

dgarske (commented):
## Skoll Code Review

**Findings:** 2 total — 2 posted, 0 skipped

### Posted findings
- [Low] **MlKem *_with_random drops buffer-protocol support for rand** — `wolfcrypt/ciphers.py:2096-2098,2140`
- [Low] **setup.py dependencies out of sync with pyproject (typing-extensions)** — `setup.py:62-63`

dgarske (changes requested):
Please resolve merge conflicts. Also see https://github.com/wolfSSL/wolfcrypt-py/pull/125#pullrequestreview-4637373251

setup.py
roberthdevries: Removed the redundant `install_requires` setting in setup.py. The overall goal is to move as much configuration settings as possible into pyproject.toml.

wolfcrypt/ciphers.py
roberthdevries: Used the same approach as in `make_key_from_seed` and `sign_with_seed` in ML-DSA. First use `memoryview` to check if it follows the buffer protocol, then convert to bytes.

dgarske (commented):
## Skoll Code Review

**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [Medium] **make_key_with_random drops buffer-protocol handling, inconsistent with encapsulate_with_random** — `wolfcrypt/ciphers.py:2151`
- [Low] **New TypeError validation branches (with_random / seed) are not covered by tests** — `wolfcrypt/ciphers.py:2096-2099`
- [Low] **Random.native_object changed from settable attribute to read-only property** — `wolfcrypt/random.py:56-60`
- [Info] **Abstract properties typed - int are overridden by None class attributes** — `wolfcrypt/ciphers.py:175-177`

wolfcrypt/ciphers.py
```diff
@@ -2071,25 +2148,23 @@ def make_key_with_random(cls, mlkem_type, rand):
             :rtype: MlKemPrivate
             """
             mlkem_priv = cls(mlkem_type)
-            ret = _lib.wc_KyberKey_MakeKeyWithRandom(
-                mlkem_priv.native_object, _ffi.from_buffer(rand), len(rand)
-            )
+            ret = _lib.wc_KyberKey_MakeKeyWithRandom(mlkem_priv.native_object, rand, len(rand))
```
dgarske, line 2159: **🟠 [Medium] make_key_with_random drops buffer-protocol handling, inconsistent with encapsulate_with_random**

This PR replaces `_ffi.from_buffer(rand)` with a bare `rand` in `wc_KyberKey_MakeKeyWithRandom(mlkem_priv.native_object, rand, len(rand))`, but does NOT add the validation/normalization that its sibling `MlKemPublic.encapsulate_with_random` gained in the same PR (lines 2096-2101: `memoryview(rand)` guard + `rand = bytes(rand)`). The old `_ffi.from_buffer(rand)` accepted any buffer-protocol object (`bytes`, `bytearray`, `memoryview`); passing such objects straight into cffi now depends on cffi's narrower direct-conversion rules, so a `bytearray`/`memoryview` caller that worked before can now raise `TypeError`. Two public entry points that take a `rand` buffer for the same crypto module now behave differently. Tests only exercise `bytes` (`ref_rand_for_make_key = h2b(...)`), so this asymmetry is not caught.

**Fix:** Mirror the validation/`bytes(rand)` conversion used in `encapsulate_with_random` so the two `*_with_random` APIs accept the same input types consistently.

wolfcrypt/ciphers.py
```diff
@@ -2016,19 +2085,27 @@ def encapsulate(self, rng=None):
...
-        def encapsulate_with_random(self, rand):
+        def encapsulate_with_random(self, rand: bytes) -> tuple[bytes, bytes]:
             """
             :param rand: random number for an encapsulation
             :type rand: bytes
             :return: tuple of a shared secret (first element) and the cipher text (second element)
             :rtype: tuple[bytes, bytes]
             """
+
+            try:
```
dgarske, line 2096: **🔵 [Low] New TypeError validation branches (with_random / seed) are not covered by tests**

The PR adds a `memoryview(rand)` buffer-protocol guard that raises a custom `TypeError` in `encapsulate_with_random` (and analogous refactors normalize `seed` in `make_key_from_seed`/`sign_with_seed`). The tests exercise only the happy path with `bytes`; the new `TypeError` branch (e.g. passing a non-buffer such as an `int` or `str`) is not tested, so a regression in that guard would go unnoticed. Combined with the finding above, the divergence between `make_key_with_random` and `encapsulate_with_random` is also untested.

**Fix:** Add a small negative test for the new buffer-protocol validation branches.

wolfcrypt/random.py
```diff
@@ -34,31 +34,37 @@ class Random:
...
-        if self.native_object:
+        if self._native_object is not None:
             try:
-                self._delete(self.native_object)
+                self._delete(self._native_object)
             except AttributeError:
                 # Can occur during interpreter shutdown
                 pass
 
+    @property
```
dgarske, line 56: **🔵 [Low] Random.native_object changed from settable attribute to read-only property**

`native_object` (a non-underscore, effectively public attribute) is converted from a plain writable attribute to a read-only `@property` backed by `_native_object`. Any external code that assigned `rng.native_object = ...` now fails with `AttributeError` (this is exactly why the test `test_delete_descriptor_binding.py` had to drop its `r.native_object = None` line). The change is otherwise reasonable and the internal `__del__`/init-failure handling is now safer, but this is a backward-incompatible surface change on a public name that is not called out in the ChangeLog (which only mentions 'Add typing annotations').

**Fix:** Confirm no external consumer sets `Random.native_object`; consider a brief ChangeLog note that it is now read-only. No code change required.

wolfcrypt/ciphers.py
```diff
@@ -152,10 +162,35 @@ def __init__(self, key, mode, IV=None):
...
+
+    @property
+    @abstractmethod
+    def _native_type(self) -> str: ...
+
+    @property
+    @abstractmethod
+    def block_size(self) -> int: ...
+
+    @property
```
dgarske, line 175: **⚪ [Info] Abstract properties typed - int are overridden by None class attributes**

In this typing-focused PR, several abstract properties are declared to return `int` (`_Cipher.key_size`, `_Cipher.block_size`, `_Cipher._key_sizes`; `_Hash.digest_size`) but concrete subclasses override them with class attributes set to `None` (`Aes.key_size = None`, `ChaCha.key_size = None`, `Sha3.digest_size = None`). This is fine at runtime (ABC only checks the name exists, and the falsy `None` is intentionally used as a sentinel in `_Cipher.__init__`), but it is type-inconsistent with the abstract signatures and will produce warnings under the newly added `[tool.ty]` config (`all = "warn"`). Consider annotating these as `int | None` (or `Optional[int]`) to match reality.

**Fix:** Widen the abstract property annotations to `int | None` where subclasses legitimately use `None` as a sentinel, to keep the type declarations consistent.

dgarske (changes requested):
https://github.com/wolfSSL/wolfcrypt-py/pull/125#pullrequestreview-4639663436

wolfcrypt/ciphers.py
roberthdevries: Fixed type annotation for `_Hash.digest_size` and `_Cipher.key_size` to `int | None`.

wolfcrypt/ciphers.py
roberthdevries: Fixed in a similar manner using `memoryview` as in `encapsulate_with_random` on line 2088.

wolfcrypt/random.py
roberthdevries: Added a note to the ChangeLog.

wolfcrypt/ciphers.py
roberthdevries: Added bad type tests for:
* `MlKemPublic.encapsulate_with_random`
* `MlKemPrivate.make_key_with_random`
* `MlDsaPrivate.make_key_from_seed`
* `MlDsaPrivate.sign_with_seed`

dgarske (approved):

PR #126, opened 2026-06-04
Prepare for v5.9.2 release

mjdemilliano:


scripts/build_ffi.py
```diff
@@ -376,6 +376,7 @@ def get_features(local_wolfssl, features):
     features["RSA_PSS"] = 1 if '#define WC_RSA_PSS' in defines else 0
     features["CHACHA20_POLY1305"] = 1 if ('#define HAVE_CHACHA' in defines and '#define HAVE_POLY1305' in defines) else 0
     features["ML_DSA"] = 1 if '#define HAVE_DILITHIUM'  in defines else 0
+    features["ML_DSA_NO_CTX"] = 1 if '#define WOLFSSL_DILITHIUM_NO_CTX' in defines else 0
```
sebastian-carpenter: MEDIUM-1: ML_DSA_NO_CTX detection misses indirect define via WOLFSSL_DILITHIUM_FIPS204_DRAFT [SUGGEST] (bug)
  File:       scripts/build_ffi.py:379
  Function:   get_features
  Confidence: Medium

  The new feature detection only checks for a literal `#define WOLFSSL_DILITHIUM_NO_CTX` in `options.h`/`user_settings.h`. However, `lib/wolfssl/wolfssl/wolfcrypt/dilithium.h` (lines 808-811) defines `WOLFSSL_DILITHIUM_NO_CTX` *internally* whenever `WOLFSSL_DILITHIUM_FIPS204_DRAFT` is enabled. In that configuration the legacy no-context functions (`wc_dilithium_sign_msg`, `wc_dilithium_sign_msg_with_seed`, `wc_dilithium_verify_msg`) are actually compiled into the library and available, but `get_features` will report `ML_DSA_NO_CTX = 0`. The cdef will then omit those declarations and `MlDsaPrivate.sign(msg)` / `verify(sig, msg)` with `ctx=None` will raise `WolfCryptError("support ... without context is disabled")` even though the build supports it. The failure is fail-safe (a clean error rather than a crash), so the impact is a degraded-functionality false negative on FIPS-204-draft builds, not a hard bug.

  Code:
    features["ML_DSA_NO_CTX"] = 1 if '#define WOLFSSL_DILITHIUM_NO_CTX' in defines else 0

  Recommendation: Also treat `WOLFSSL_DILITHIUM_FIPS204_DRAFT` as implying no-ctx support, mirroring the logic in dilithium.h, so the legacy API is exposed in those builds.

tests/test_mldsa.py
sebastian-carpenter, line 114: MEDIUM-2: No test exercises the new 'context disabled' raise path [SUGGEST] (test)
  File:       tests/test_mldsa.py:113-160
  Function:   test_sign_verify
  Confidence: High

  The diff introduces three new branches in `ciphers.py` (verify/sign/sign_with_seed) that raise `WolfCryptError("support ... without context is disabled")` when `ctx is None` and `_lib.ML_DSA_NO_CTX` is false. The tests only ever take the ctx=None path inside `if _lib.ML_DSA_NO_CTX:` guards or `@pytest.mark.skipif(not _lib.ML_DSA_NO_CTX)`, so when running against a build with no-ctx disabled (the likely default for the bumped wolfSSL 5.9.1 with just `--enable-dilithium`), the new raise branches are never asserted. The behavior of the new error path is untested.

  Code:
    elif not _lib.ML_DSA_NO_CTX:
        raise WolfCryptError("support for signing without context is disabled")

  Recommendation: Add an assertion (guarded by `if not _lib.ML_DSA_NO_CTX:`) that signing/verifying with `ctx=None` raises `WolfCryptError`, so the new branch is covered in no-ctx builds.

scripts/build_ffi.py
sebastian-carpenter, line 1390: LOW-3: Default features dict omits ML_DSA_NO_CTX while listing peer flags [NIT] (convention)
  File:       scripts/build_ffi.py:1348-1378
  Function:   main
  Confidence: High

  The default `features` dictionary lists `ML_DSA`, `ML_KEM`, `HKDF`, etc., but the newly introduced `ML_DSA_NO_CTX` key is not present. In practice this is harmless because `get_features()` (which always sets `ML_DSA_NO_CTX` unconditionally at line 379) is always called before `build_ffi()` in `main()`, so no `KeyError` occurs. However, it is inconsistent with how the other ML-DSA/ML-KEM flags are seeded and would break if `build_ffi` were ever invoked without `get_features` first.

  Code:
    "ML_KEM": 1,
    "ML_DSA": 1,
    "HKDF": 1,

  Recommendation: Add `"ML_DSA_NO_CTX": 0` to the default features dict for consistency and robustness.

wolfcrypt/ciphers.py
```diff
@@ -2264,6 +2264,8 @@ def verify(self, signature, message, ctx=None):
                 )
                 if ret < 0:  # pragma: no cover
                     raise WolfCryptApiError("wc_dilithium_verify_ctx_msg() error", ret)
+            elif not _lib.ML_DSA_NO_CTX:
+                raise WolfCryptError("support verifying without context is disabled")
```
sebastian-carpenter: LOW-4: Inconsistent wording between sign and verify 'disabled' error messages [NIT] (style)
  File:       wolfcrypt/ciphers.py:2268
  Function:   MlDsaPublic.verify
  Confidence: High

  The verify path raises `"support verifying without context is disabled"`, while the two sign paths raise `"support for signing without context is disabled"`. The verify message is missing the word "for" and reads awkwardly compared to the sign messages.

  Code:
    elif not _lib.ML_DSA_NO_CTX:
        raise WolfCryptError("support verifying without context is disabled")

  Recommendation: Align the verify message wording with the sign messages for consistency.

wolfcrypt/ciphers.py
sebastian-carpenter, line 2243: LOW-5: Docstrings still state ctx=None means 'no context' but it can now raise [NIT] (api)
  File:       wolfcrypt/ciphers.py:2238-2248,2410-2420,2464-2474
  Function:   MlDsaPublic.verify / MlDsaPrivate.sign / MlDsaPrivate.sign_with_seed
  Confidence: Medium

  The `ctx` parameter docstrings still say ":type ctx: None for no context, str or bytes otherwise". After this diff, passing `ctx=None` raises `WolfCryptError` on builds where `WOLFSSL_DILITHIUM_NO_CTX` is not compiled in (likely the default for the bumped wolfSSL 5.9.1). The documentation no longer reflects that `ctx=None` is conditionally supported, which can confuse callers who follow the docs and hit an exception. Note FIPS-204-compliant empty-context behavior is still reachable via `ctx=b""`, which is distinct from the legacy `ctx=None` path.

  Code:
    :param ctx: context (optional)
    :type ctx: None for no context, str or bytes otherwise

  Recommendation: Update the docstrings to note that `ctx=None` requires no-ctx support and that `ctx=b""` provides FIPS-204 empty-context signing/verification.

tests/test_aesgcmstream.py
```diff
@@ -141,7 +141,14 @@ def test_invalid_tag_bytes():
...
             gcm.encrypt("hello world")
-            tag = gcm.final()
+            try:
+                tag = gcm.final()
+            except WolfCryptApiError as exception:
+                if "(-173)" in str(exception) and good < 12:
+                    # Probably this authentication tag size is not supported,
+                    # the minimum is 12 by default. Ignore.
+                    continue
+                raise
```
sebastian-carpenter: gate this on WOLFSSL_MIN_AUTH_TAG_SZ. This try-except loop is too ambiguous.

sebastian-carpenter (changes requested):

scripts/build_ffi.py
mjdemilliano: Done

tests/test_mldsa.py
mjdemilliano: Done

wolfcrypt/ciphers.py
mjdemilliano: Done

tests/test_aesgcmstream.py
mjdemilliano: Done

mjdemilliano (commented):
Processed all comments

mjdemilliano:
Would like to include #129 , so marking as draft for now

tests/test_aesgcmstream.py
```diff
@@ -139,16 +139,11 @@ def test_invalid_tag_bytes():
                 AesGcmStream(key, iv, tag_bytes=bad)
         # Valid NIST sizes: verify the resulting tag has the requested length.
         for good in (4, 8, 12, 13, 14, 15, 16):
+            if good < _lib.MIN_AUTH_TAG_SZ:
+                continue
```
sebastian-carpenter: Just skipping the size like this loses coverage. In the previous review all these sizes were tested, with an error expected for some. I'd like to preserve that - but expect a general error for sizes below the threshold rather than a literal numeric code.

scripts/build_ffi.py
```diff
@@ -540,6 +547,7 @@ def build_ffi(local_wolfssl, features):
         extern int ML_DSA_ENABLED;
         extern int ML_DSA_NO_CTX;
```
sebastian-carpenter: ### INFO-2: Boolean flag ML_DSA_NO_CTX omits the _ENABLED suffix used by sibling flags
- **File:** `scripts/build_ffi.py:506,548`
- **Function:** `build_ffi`
- **Action:** NIT
- **Tag:** convention
- **Confidence:** Medium

**Description:** `ML_DSA_NO_CTX` is a 0/1 boolean capability flag exposed via `_lib`, but every other boolean capability flag in this list uses the `_ENABLED` suffix (`ML_DSA_ENABLED`, `ML_KEM_ENABLED`, `AESGCM_STREAM_ENABLED`, `WC_RNG_SEED_CB_ENABLED`, etc.). The new flag breaks that pattern. (`MIN_AUTH_TAG_SZ` is a numeric value, so it correctly does not take the suffix.)

**Code:**
```
int ML_DSA_NO_CTX = {features["ML_DSA_NO_CTX"]};
...
extern int ML_DSA_NO_CTX;
```

**Recommendation:** Optional rename for convention consistency; not blocking.

tests/test_mldsa.py
sebastian-carpenter, line 178: MEDIUM-1: sign_with_seed determinism coverage lost on default (non-NO_CTX) builds [SUGGEST] (test)
  File:       tests/test_mldsa.py:171-199
  Function:   test_sign_with_seed
  Confidence: Medium

  The diff adds `@pytest.mark.skipif(not _lib.ML_DSA_NO_CTX, ...)` to `test_sign_with_seed`. Confirmed from `lib/wolfssl/configure.ac:1822-1823`, `WOLFSSL_DILITHIUM_NO_CTX` is only defined for `--enable-dilithium=no-ctx` or the FIPS204 draft option; the plain `--enable-dilithium` used by `scripts/build_ffi.py:239` does NOT define it. So on the default wolfcrypt-py build `_lib.ML_DSA_NO_CTX == 0` and this entire test is skipped. The skipped test is the ONLY place that asserts deterministic signing (`sign_with_seed(message, seed)` twice → identical signatures, lines 190-191) and the only place testing the seed-length/seed-type validation (lines 194-199). The sibling `test_sign_with_seed_and_context` (not skipped) signs once with `ctx=` but never re-signs to assert determinism, so determinism of `sign_with_seed` — a core property of the newly-added deterministic-signing feature — has zero coverage on the most common build configuration. Determinism does not actually require no-context support; it could be re-asserted by passing a `ctx=` value.

  Code:
        @pytest.mark.skipif(not _lib.ML_DSA_NO_CTX, reason="Requires support for signing without context")
        def test_sign_with_seed(mldsa_type, rng):
            ...
            signature = mldsa_priv.sign_with_seed(message, signature_seed)
            ...
            # re-generate from the same seed:
            signature_from_same_seed = mldsa_priv.sign_with_seed(message, signature_seed)
            assert signature == signature_from_same_seed

  Recommendation: Add a determinism assertion (re-sign with the same seed and compare) to `test_sign_with_seed_and_context`, which already runs on default builds, so the deterministic-signing guarantee is exercised regardless of NO_CTX support.

sebastian-carpenter, line 114: MEDIUM-1: No test coverage for the newly-documented empty-context (ctx=b"") ML-DSA path [SUGGEST] (test)
  File:       tests/test_mldsa.py:146-169
  Function:   test_sign_verify
  Confidence: Medium

  This diff simultaneously (a) disables the `ctx=None` no-context path on default builds (the new `elif not _lib.ML_DSA_NO_CTX: raise WolfCryptError(...)` guards in `sign`/`sign_with_seed`/`verify`) and (b) updates the docstrings to instruct users to `pass empty string "" for FIPS-204 empty-context signing/verification`. On the standard `--enable-dilithium` build (where `_lib.ML_DSA_NO_CTX == 0`), the empty-context call form is now the documented replacement for the disabled `None` default. However, no test exercises it. The existing tests only cover non-empty `ctx=ctx`, `ctx=wrong_ctx`, and (for the raise path) `ctx` omitted. The empty-context boundary (`ctx=b""` → `t2b("")` → `b""` → `wc_dilithium_sign_ctx_msg(..., ctxLen=0, ...)`) is precisely the FIPS-204 path the new docs steer default-build users toward, yet it is never executed or round-trip verified. A regression in the `len==0` ctx handling (or a wolfSSL change to it) would pass CI silently.

  Code:
            # Sign a message with context
            signature = mldsa_priv.sign(message, rng, ctx=ctx)
            ...
            if not _lib.ML_DSA_NO_CTX:
                with pytest.raises(WolfCryptError):
                    mldsa_priv.sign(message)
                ...
            # Verify with wrong context
            assert not mldsa_pub.verify(signature, message, ctx=wrong_ctx)

  Recommendation: Add an explicit empty-context (`ctx=b""`) sign+verify round-trip assertion to `test_sign_verify` so the documented FIPS-204 empty-context behavior — the de-facto default usage on standard builds after this change — is actually exercised.

sebastian-carpenter (changes requested):

tests/test_aesgcmstream.py
mjdemilliano: Please review the changes I made to address your comment.

tests/test_mldsa.py
mjdemilliano: Done.

mjdemilliano: Done.

dgarske (changes requested):
Please update to v5.9.2 that was tagged last week. Thanks!

tests/test_aesgcmstream.py
sebastian-carpenter: That looks good. I think this finding will probably be the last:

  MEDIUM-1: New MIN_AUTH_TAG_SZ constructor guard couples unchanged test_encrypt_short_tag to the build configuration [SUGGEST] (test)
  File:       tests/test_aesgcmstream.py:47-58
  Function:   test_encrypt_short_tag
  Confidence: Medium

  This diff adds `if tag_bytes < _lib.MIN_AUTH_TAG_SZ: raise ValueError(...)` to the `AesGcmStream.__init__` constructor (`wolfcrypt/ciphers.py:407-411`). The sibling test `test_invalid_tag_bytes` was updated to account for this (it wraps small tag sizes in `pytest.raises`), but `test_encrypt_short_tag` was NOT — it hardcodes `AesGcmStream(key, iv, 12)` with no guard. On the standard wolfcrypt-py build `MIN_AUTH_TAG_SZ` defaults to 12, so `12 < 12` is False and the test still passes. However, the new constructor check now couples this previously build-agnostic test to the compiled `WOLFSSL_MIN_AUTH_TAG_SZ`: on any build where `MIN_AUTH_TAG_SZ > 12`, `test_encrypt_short_tag` would raise `ValueError` at construction and fail, whereas before the diff it worked on any AESGCM_STREAM-capable build. The diff therefore worsened the portability of this unchanged test. This is distinct from the known-issue note about `test_invalid_tag_bytes`.

  Code:
        def test_encrypt_short_tag():
            key = "fedcba9876543210"
            iv = "0123456789abcdef"
            gcm = AesGcmStream(key, iv, 12)   # <-- now coupled to _lib.MIN_AUTH_TAG_SZ
            ...
            gcmdec = AesGcmStream(key, iv, 12)

  Recommendation: Guard `test_encrypt_short_tag` with a `pytest.mark.skipif(12 < _lib.MIN_AUTH_TAG_SZ, ...)` (mirroring how `test_invalid_tag_bytes` was updated) so it remains correct on builds that raise the minimum auth-tag size above 12. On the default build it is currently fine.

tests/test_aesgcmstream.py
sebastian-carpenter, line 49: MEDIUM-1: New MIN_AUTH_TAG_SZ constructor guard couples unchanged test_encrypt_short_tag to the build configuration [SUGGEST] (test)
  File:       tests/test_aesgcmstream.py:47-58
  Function:   test_encrypt_short_tag
  Confidence: Medium

  This diff adds `if tag_bytes < _lib.MIN_AUTH_TAG_SZ: raise ValueError(...)` to the `AesGcmStream.__init__` constructor (`wolfcrypt/ciphers.py:407-411`). The sibling test `test_invalid_tag_bytes` was updated to account for this (it wraps small tag sizes in `pytest.raises`), but `test_encrypt_short_tag` was NOT — it hardcodes `AesGcmStream(key, iv, 12)` with no guard. On the standard wolfcrypt-py build `MIN_AUTH_TAG_SZ` defaults to 12, so `12 < 12` is False and the test still passes. However, the new constructor check now couples this previously build-agnostic test to the compiled `WOLFSSL_MIN_AUTH_TAG_SZ`: on any build where `MIN_AUTH_TAG_SZ > 12`, `test_encrypt_short_tag` would raise `ValueError` at construction and fail, whereas before the diff it worked on any AESGCM_STREAM-capable build. The diff therefore worsened the portability of this unchanged test. This is distinct from the known-issue note about `test_invalid_tag_bytes`.

  Code:
        def test_encrypt_short_tag():
            key = "fedcba9876543210"
            iv = "0123456789abcdef"
            gcm = AesGcmStream(key, iv, 12)   # <-- now coupled to _lib.MIN_AUTH_TAG_SZ
            ...
            gcmdec = AesGcmStream(key, iv, 12)

  Recommendation: Guard `test_encrypt_short_tag` with a `pytest.mark.skipif(12 < _lib.MIN_AUTH_TAG_SZ, ...)` (mirroring how `test_invalid_tag_bytes` was updated) so it remains correct on builds that raise the minimum auth-tag size above 12. On the default build it is currently fine.

sebastian-carpenter (changes requested):

tests/test_aesgcmstream.py
mjdemilliano: I don't really understand the purpose/added value of this test but I tried to just implement the suggestion.

mjdemilliano:
> Please update to v5.9.2 that was tagged last week. Thanks!

Done

scripts/build_ffi.py
```diff
@@ -373,9 +373,16 @@ def get_features(local_wolfssl, features):
     features["ASN"] = 0 if '#define NO_ASN' in defines else 1
     features["WC_RNG_SEED_CB"] = 1 if '#define WC_RNG_SEED_CB' in defines else 0
     features["AESGCM_STREAM"] = 1 if '#define WOLFSSL_AESGCM_STREAM' in defines else 0
+    # Try to read minimum AESGCM authentication tag size from settings, else use default.
+    min_auth_tag_sz = re.search(r'#define\s+WOLFSSL_MIN_AUTH_TAG_SZ\s+(\d+)', '\n'.join(defines))
+    features["MIN_AUTH_TAG_SZ"] = int(min_auth_tag_sz.group(1)) if min_auth_tag_sz else 12
     features["RSA_PSS"] = 1 if '#define WC_RSA_PSS' in defines else 0
     features["CHACHA20_POLY1305"] = 1 if ('#define HAVE_CHACHA' in defines and '#define HAVE_POLY1305' in defines) else 0
     features["ML_DSA"] = 1 if '#define HAVE_DILITHIUM'  in defines else 0
+    features["ML_DSA_NO_CTX"] = 1 if (
```
dgarske: **🟠 [Medium] ML_DSA_NO_CTX detection misses newer canonical WOLFSSL_MLDSA_* macro names**

The new feature detection only text-matches the legacy `WOLFSSL_DILITHIUM_NO_CTX` and `WOLFSSL_DILITHIUM_FIPS204_DRAFT` macros in `options.h`/`user_settings.h`:
```python
features["ML_DSA_NO_CTX"] = 1 if (
    '#define WOLFSSL_DILITHIUM_NO_CTX' in defines
    or '#define WOLFSSL_DILITHIUM_FIPS204_DRAFT' in defines
) else 0
```
In wolfSSL 5.9.2 (commit ac01707) the no-ctx legacy functions (`wc_dilithium_sign_msg`, `wc_dilithium_sign_msg_with_seed`, `wc_dilithium_verify_msg`) are actually gated on `WOLFSSL_MLDSA_NO_CTX`, which the headers treat as the new canonical name. `dilithium.h` maps the names both directions, but only at C-preprocessor time. The build script does a raw text search of the header text, so a custom build that defines the newer `WOLFSSL_MLDSA_NO_CTX` (or `WOLFSSL_MLDSA_FIPS204_DRAFT`) directly in `user_settings.h` will NOT be detected. Result: the C library exposes the no-ctx functions, but `ML_DSA_NO_CTX_ENABLED` is set to 0, the cdef omits the declarations, and `MlDsaPrivate.sign(msg)` / `verify(...)` with `ctx=None` raise `WolfCryptError("support ... disabled")` even though the build supports it. The standard wolfcrypt-py release build (`--enable-dilithium` only) is unaffected because no-ctx is off by default, so this is a forward-compat/custom-build robustness gap, not a release blocker.

**Fix:** Also accept the newer `WOLFSSL_MLDSA_NO_CTX` and `WOLFSSL_MLDSA_FIPS204_DRAFT` macro names in the detection (ideally via a regex tolerant of whitespace/values, as was done for MIN_AUTH_TAG_SZ) so custom builds using the canonical macro names are recognized.

tests/test_aesgcmstream.py
```diff
@@ -42,7 +45,9 @@ def test_encrypt():
         gcmdec.final(authTag)
         assert bufdec == t2b("hello world")
 
+    @pytest.mark.skipif(12 < _lib.MIN_AUTH_TAG_SZ, reason="test can only be performed when MIN_AUTH_TAG_SZ is larger than 12")
```
dgarske: **🔵 [Low] skipif reason string for test_encrypt_short_tag is logically inverted**

The skip condition is `12 < _lib.MIN_AUTH_TAG_SZ`, i.e. the test (which exercises a 12-byte tag) is SKIPPED when the build's minimum auth tag size is greater than 12. The `reason` text states the opposite: "test can only be performed when MIN_AUTH_TAG_SZ is larger than 12". When the test is skipped, the real reason is that a 12-byte tag is NOT supported because the minimum is larger than 12. The displayed reason is misleading to anyone reading skip output.

**Fix:** Reword the reason string to match the condition (test runs only when MIN_AUTH_TAG_SZ <= 12).

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **ML_DSA_NO_CTX detection misses newer canonical WOLFSSL_MLDSA_* macro names** — `scripts/build_ffi.py:382-385`
- [Low] **skipif reason string for test_encrypt_short_tag is logically inverted** — `tests/test_aesgcmstream.py:48`
- [Info] **Branch name (prepare-v5.9.1) does not match the v5.9.2 release content** — `wolfcrypt/_version.py:3-11`

wolfcrypt/_version.py
```diff
@@ -1,12 +1,12 @@
 # When bumping the C library version, reset the POST count to 0
 
-__wolfssl_version__ = "v5.8.4-stable"
+__wolfssl_version__ = "v5.9.2-stable"
```
dgarske, line 3: **⚪ [Info] Branch name (prepare-v5.9.1) does not match the v5.9.2 release content**

The PR bumps `__wolfssl_version__` to `v5.9.2-stable`, `__version__` to `5.9.2-0`, and the ChangeLog header to `Release 5.9.2 (Jul 1, 2026)` — all internally consistent for a 5.9.2 release. However the PR head branch is named `prepare-v5.9.1`, which is stale/incorrect for a 5.9.2 release. Purely a metadata/hygiene observation; no code impact.

**Fix:** Optionally rename the branch to reflect v5.9.2 to avoid confusion; no code change required.

scripts/build_ffi.py
mjdemilliano: Done. The regex thing could be applied in more places but that is out of scope for me, and should not impede this release.

tests/test_aesgcmstream.py
mjdemilliano: Fixed

mjdemilliano:
Pushed exact same changes as #134 so I propose to continue the discussion there with the new branch name.

PR #127, opened 2026-06-21
Enable more ruff rules requiring only minor or no fixes.

roberthdevries:
Extra rules:
All E pycodestyle error rules
C4   flake8-comprehensions
DTZ  flake8-datetimez
EXE  flake8-executable
FA   flake8-future-annotations
INT  flake8-gettext
ISC  flake8-implicit-str-concat
ICN  flake8-import-conventions
LOG  flake8-logging
G    flake8-logging-format
RSE  flake8-raise
SLOT flake8-slots
TID  flake8-tidy-imports
TC   flake8-type-checking
FLY  flynt
PERF Perflint
W    pycodestyle warnings
FURB refurb

Ignore E501:  line too long

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,9 +96,6 @@ indent-width = 4
 target-version = "py310"
 
 [tool.ruff.lint]
-# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
-# Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
-# McCabe complexity (`C901`) by default.
 select = ["E", "F", "B", "UP", "C4", "DTZ", "EXE", "FA", "INT", "ISC", "ICN", "LOG", "G", "RSE", "SLOT", "TID", "TC", "FLY", "PERF", "W", "FURB"]
 ignore = ["E501"]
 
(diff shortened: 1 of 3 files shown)


dgarske:
Okay to test. Contributor agreement on file.

pyproject.toml
```diff
@@ -99,8 +99,8 @@ target-version = "py310"
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
```
dgarske: **🔵 [Low] Stale comment above ruff select no longer matches the configuration**

The comment block preceding `select` still reads: `Enable Pyflakes (F) and a subset of the pycodestyle (E) codes by default. Unlike Flake8, Ruff doesn't enable pycodestyle warnings (W)...`. After this change, `select` now enables the full `E` family (not 'a subset') and explicitly enables pycodestyle warnings `W`, so the comment now contradicts the actual config. This is left over from the previous `select = ["E4", "E7", "E9", ...]` line and could mislead a future reader.

**Fix:** Update the comment to reflect that the full `E` and `W` families are now enabled, or drop the now-inaccurate sentence. Non-blocking.

dgarske (commented):
## Skoll Code Review

**Findings:** 1 total — 1 posted, 0 skipped

### Posted findings
- [Low] **Stale comment above ruff select no longer matches the configuration** — `pyproject.toml:99-103`

pyproject.toml
roberthdevries: Removed the now obsolete comment block.

dgarske (approved):

PR #130, opened 2026-06-26
Random: add DRBG reseed support

mjdemilliano:
Only available when Hash-DRBG is enabled, which is the default.


tests/test_random.py
```diff
@@ -48,3 +50,17 @@ def test_nonce_byte(rng_nonce):
...
     assert len(rng_nonce.bytes(length)) == length
+
+
+@pytest.mark.skipif(not _lib.HASHDRBG_ENABLED, reason="Reseeding only available with hash-DRBG")
+def test_reseed(rng):
+    rng.reseed(b"some seed material for testing")
+    assert len(rng.bytes(32)) == 32
+
+
+@pytest.mark.skipif(not _lib.HASHDRBG_ENABLED, reason="Reseeding only available with hash-DRBG")
```
dgarske, line 71: **🟠 [Medium] test_reseed_empty asserts nothing and silently passes on either outcome**

The test wraps `rng.reseed(b"")` in a try/except that swallows `WolfCryptApiError` with `pass`, and has no assertion. This means the test passes whether the C layer rejects the empty seed (raises) OR accepts it (returns success). It therefore verifies no observable behavior and would not catch a regression where `reseed` crashes the interpreter short of raising, or where it returns a bogus positive value. If the intent is to document that an empty seed is rejected, use `pytest.raises`; if empty seed is genuinely allowed, assert subsequent generation still works. As written it is effectively a smoke test that the call does not hard-crash.

**Fix:** Decide the expected contract for an empty seed and assert it explicitly (use pytest.raises if rejection is the contract), rather than passing unconditionally.

dgarske (commented):
## Skoll Code Review

**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [Medium] **test_reseed_empty asserts nothing and silently passes on either outcome** — `tests/test_random.py:61-66`
- [Low] **Non-ASCII em-dash in test comment violates ASCII-only convention** — `tests/test_random.py:66`
- [Low] **test_reseed only checks output length, not that reseed had any effect** — `tests/test_random.py:55-58`
- [Low] **Per-method feature gate inside class body differs from repo's module-level gating pattern** — `wolfcrypt/random.py:81-89`

tests/test_random.py
```diff
@@ -48,3 +50,17 @@ def test_nonce_byte(rng_nonce):
...
+def test_reseed_empty(rng):
+    try:
+        rng.reseed(b"")
+    except WolfCryptApiError:
+        pass  # acceptable — C rejects zero-length seed
```
dgarske: **🔵 [Low] Non-ASCII em-dash in test comment violates ASCII-only convention**

The comment `pass  # acceptable — C rejects zero-length seed` contains a U+2014 em-dash, which is not 7-bit ASCII. The project mandates ASCII-only source, comments, and commit messages (no smart quotes, em/en dashes, ellipses, or arrows). The rest of the file uses plain ASCII. Replace the em-dash with `-` or `--`.

**Fix:** Replace the em-dash with an ASCII hyphen to comply with the repo's ASCII-only convention.

tests/test_random.py
```diff
@@ -48,3 +50,17 @@ def test_nonce_byte(rng_nonce):
 @pytest.mark.parametrize("length", (1, 8, 128))
 def test_nonce_bytes(rng_nonce, length):
     assert len(rng_nonce.bytes(length)) == length
+
+
+@pytest.mark.skipif(not _lib.HASHDRBG_ENABLED, reason="Reseeding only available with hash-DRBG")
```
dgarske, line 57: **🔵 [Low] test_reseed only checks output length, not that reseed had any effect**

test_reseed verifies that `reseed()` returns without error and that a subsequent `bytes(32)` produces 32 bytes, but `bytes(32)` already has length coverage via test_bytes. The new test adds little beyond confirming reseed does not raise. Consider also exercising that generation continues to succeed across multiple reseed calls, or reseeding with differing seed material, to give the new binding more meaningful coverage. This is optional given that binding-level tests mostly confirm the FFI call wiring.

**Fix:** Optionally strengthen the happy-path test to cover repeated reseeds; not blocking.

wolfcrypt/random.py
```diff
@@ -77,3 +77,13 @@ def bytes(self, length: int) -> __builtins__.bytes:
             raise WolfCryptApiError("RNG generate block error", ret)
 
         return _ffi.buffer(result, length)[:]
+
+    if _lib.HASHDRBG_ENABLED:
```
dgarske, line 81: **🔵 [Low] Per-method feature gate inside class body differs from repo's module-level gating pattern**

Elsewhere in the codebase, feature flags gate whole classes/functions at module scope (e.g. ciphers.py `if _lib.RSA_ENABLED:`, hkdf.py `if _lib.HKDF_ENABLED:`, pwdbased.py `if _lib.PWDBASED_ENABLED:`). Here the gate `if _lib.HASHDRBG_ENABLED:` wraps a single method inside the class body, which is valid Python but a new pattern for this repo. A consequence is that when HASHDRBG is disabled, `Random` instances simply lack `reseed` and callers get an `AttributeError` rather than a descriptive message. This is consistent with how absent gated classes behave, so it is acceptable; flagging only for awareness. If a clearer failure mode is desired, define `reseed` unconditionally and raise a descriptive error when `not _lib.HASHDRBG_ENABLED`.

**Fix:** Acceptable as-is; optionally define reseed unconditionally and raise a clear "reseed requires Hash-DRBG" error when the feature is disabled for a friendlier API.

tests/test_random.py
mjdemilliano: I agree with the comment. Decided to allow for an empty seed, because apparently the C API also allows this.

tests/test_random.py
mjdemilliano: No longer applicable.

tests/test_random.py
mjdemilliano: Done

wolfcrypt/random.py
mjdemilliano: I understand the point, but I would argue that this pattern happens a lot in this repo, e.g. in the `RsaPrivate` class and `EccPublic`. So I believe this is consistent. Also, I don't see any example where an appropriate exception is raised, so I wouldn't know how to do that in a consistent way. I would therefore keep this as it is.

dgarske (commented):
## Skoll Multi-Scan Review

**Modes:** `review + review-security`**Overall recommendation:** `COMMENT`

### Posted findings
- [Medium] [review] **reseed error branch missing # pragma: no cover used by every other error path** — `wolfcrypt/random.py:88`
- [Low] [review] **Trailing whitespace on blank line inside test_reseed_multiple** — `tests/test_random.py:79`
- [Low] [review] **Seed bytes use % 255, likely intended % 256** — `tests/test_random.py:61,77`

wolfcrypt/random.py
```diff
@@ -77,3 +77,13 @@ def bytes(self, length: int) -> __builtins__.bytes:
...
+        def reseed(self, seed: __builtins__.bytes) -> None:
+            """
+            Reseed the DRBG with the provided seed material.
+            """
+            assert self.native_object is not None
+            ret = _lib.wc_RNG_DRBG_Reseed(self.native_object, seed, len(seed))
+            if ret < 0:
```
dgarske: **🟠 [Medium] reseed error branch missing # pragma: no cover used by every other error path** · _convention_

The new `reseed` method's error branch `if ret < 0:` does not carry the `# pragma: no cover` marker that every other native-error branch in the codebase uses, including the three sibling methods in this same file (`random.py:40`, `:63`, `:76`) and all of `ciphers.py`/`hashes.py`. The reseed success path is exercised by the new tests, but the failure path is not (and there is no easy way to force `wc_RNG_DRBG_Reseed` to fail from Python). This leaves the only un-annotated uncovered branch in the module. CI does not currently enforce a coverage threshold (tox runs `py.test tests/` with no `--cov-fail-under`), so this will not break the build, but it is an inconsistency with the established convention and would lower the reported coverage for this file.

**Fix:** Add `# pragma: no cover` to the `if ret < 0:` line to match the convention applied to all other error branches in the package.

tests/test_random.py
```diff
@@ -48,3 +49,34 @@ def test_nonce_byte(rng_nonce):
...
+    """
+    Test that consecutive reseeding of the random number generator works.
+    """
+    # Using our own rng for getting random seed sizes.
+    for _ in range(10):
+        # Create seed using random seed size for each call.
+        seed_size = ord(rng.byte())
+        seed = bytes(x % 255 for x in range(seed_size))
+        rng.reseed(seed)
+
```
dgarske: **🔵 [Low] Trailing whitespace on blank line inside test_reseed_multiple** · _style_

Line 79 is a blank line that contains four trailing spaces. The project's `ruff` lint selection (`E4,E7,E9,F,B,UP`) does not include the pycodestyle `W` codes and the CI flake8 step is commented out, so this will not fail CI, but `ruff format` would strip it and it is a stray whitespace artifact introduced by this PR.

**Fix:** Remove the trailing whitespace on the blank line (run `ruff format tests/test_random.py`).

tests/test_random.py
```diff
@@ -48,3 +49,34 @@ def test_nonce_byte(rng_nonce):
...
+
+@pytest.mark.skipif(not _lib.HASHDRBG_ENABLED, reason="Reseeding only available with hash-DRBG")
+@pytest.mark.parametrize("seed_size", [0, 1, 32, 1000])
+def test_reseed_sizes(rng, seed_size):
+    """
+    Test that reseeding the random number generator works, for various seed sizes.
+    """
+    # Create seed of required length.
+    seed = bytes(x % 255 for x in range(seed_size))
```
dgarske: **🔵 [Low] Seed bytes use % 255, likely intended % 256** · _test_

Both new tests build seed material with `bytes(x % 255 for x in range(seed_size))`. Using `% 255` means the value 255 is never produced and the sequence wraps 254 -> 0, so the full byte range is not exercised. This does not affect test validity (the reseed accepts any bytes), but `% 256` is almost certainly what was intended if the goal was to cover all byte values.

**Fix:** Change `% 255` to `% 256` in both tests if full-byte-range seed data was intended; otherwise leave as-is.

wolfcrypt/random.py
mjdemilliano: Done

tests/test_random.py
mjdemilliano: Done

tests/test_random.py
mjdemilliano: Done

dgarske (approved):

scripts/build_ffi.py
```diff
@@ -378,6 +378,7 @@ def get_features(local_wolfssl, features):
     features["ML_DSA"] = 1 if '#define HAVE_DILITHIUM'  in defines else 0
     features["ML_KEM"] = 1 if '#define WOLFSSL_HAVE_MLKEM'  in defines else 0
     features["HKDF"] = 1 if "#define HAVE_HKDF" in defines else 0
+    features["HASHDRBG"] = 1 if "#define HAVE_HASHDRBG" in defines else 0
```
dgarske: **🟠 [Medium] HASHDRBG feature not detected on Windows non-FIPS build; reseed() silently unavailable**

The new detection `features["HASHDRBG"] = 1 if "#define HAVE_HASHDRBG" in defines else 0` does a literal, exact-line match against the defines pulled from `options.h`/`user_settings.h`. On Linux/macOS the autotools build emits `#define HAVE_HASHDRBG` into the generated `options.h` (verified in `lib/wolfssl/.../include/wolfssl/options.h`), so detection works. On Windows non-FIPS the build copies `windows/non_fips/user_settings.h` as the sole defines source, and that file does NOT contain `#define HAVE_HASHDRBG` — Hash-DRBG is left enabled implicitly via wolfSSL's default `settings.h` (no `WC_NO_HASHDRBG`). As a result, `features["HASHDRBG"]` evaluates to 0 on Windows non-FIPS even though `wc_RNG_DRBG_Reseed` is actually compiled into the library. Consequently `HASHDRBG_ENABLED` is 0, `wc_RNG_DRBG_Reseed` is omitted from the cdef, and `Random.reseed()` is never defined on the default Windows build — the very feature this PR adds ships as unavailable there. Note `windows/fips_ready/user_settings.h:17` already defines `HAVE_HASHDRBG`, and `windows/non_fips/user_settings.h:23` already lists `HAVE_HKDF` for exactly this reason, so the omission looks accidental.

**Fix:** Add `#define HAVE_HASHDRBG` to `windows/non_fips/user_settings.h` (mirroring the existing `HAVE_HKDF` define and the fips_ready variant) so the new detection fires and `reseed()` is exposed on the default Windows build. Alternatively, document that reseed is Linux/macOS-only for now.

tests/test_random.py
```diff
@@ -38,13 +39,47 @@ def test_bytes(rng):
...
+    # Create seed of required length.
+    seed = bytes(x % 256 for x in range(seed_size))
+    assert len(seed) == seed_size
+    rng.reseed(seed)
+    # Pull some bytes from the random number generator to test that it still works.
+    rng.bytes(32)
+
+
+@pytest.mark.skipif(not _lib.HASHDRBG_ENABLED, reason="Reseeding only available with hash-DRBG")
+def test_reseed_multiple(rng):
```
dgarske, line 72: **🔵 [Low] test_reseed_multiple is non-deterministic**

The test derives both the per-iteration seed size and the final byte count from live RNG output (`seed_size = ord(rng.byte())`, `num_bytes = ord(rng.byte())`). This makes the test path non-deterministic: a failure would not reproduce with a fixed input, and coverage varies run to run. Values are bounded to 0-255 so there is no resource risk, but reproducibility suffers. The success-path coverage itself is otherwise good (test_reseed_sizes already exercises 0/1/32/1000).

**Fix:** Use fixed, explicit seed sizes and byte counts so the test is deterministic and reproducible; keep the loop to prove consecutive reseeds work.

wolfcrypt/random.py
```diff
@@ -77,3 +77,13 @@ def bytes(self, length: int) -> __builtins__.bytes:
             raise WolfCryptApiError("RNG generate block error", ret)
 
         return _ffi.buffer(result, length)[:]
+
+    if _lib.HASHDRBG_ENABLED:
+        def reseed(self, seed: __builtins__.bytes) -> None:
```
dgarske, line 82: **🔵 [Low] reseed() accepts bytes only with no input normalization**

`reseed` passes `seed` straight to the CFFI call. If a caller passes a `str` instead of `bytes`, CFFI raises an opaque type error rather than a clear message. Other modules (e.g. `hkdf.py`) normalize inputs via `t2b()`. This is consistent with `Random.__init__` (which also takes `nonce` as raw bytes without `t2b`), so within `random.py` the convention is followed; flagging only for awareness since the type hint (`__builtins__.bytes`) is not enforced at runtime.

**Fix:** Optional: either leave as-is to match Random.__init__'s bytes-only convention, or run `seed` through `t2b()` for consistency with the rest of the bindings and friendlier errors on str input.

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **HASHDRBG feature not detected on Windows non-FIPS build; reseed() silently unavailable** — `scripts/build_ffi.py:381`
- [Low] **test_reseed_multiple is non-deterministic** — `tests/test_random.py:72-85`
- [Low] **reseed() accepts bytes only with no input normalization** — `wolfcrypt/random.py:82-89`

dgarske (changes requested):
https://github.com/wolfSSL/wolfcrypt-py/pull/130#pullrequestreview-4621733725

scripts/build_ffi.py
mjdemilliano: Fixed

tests/test_random.py
mjdemilliano: Fixed

wolfcrypt/random.py
mjdemilliano: Leaving as-is to keep this module consistent in being bytes-only. The method is clearly type-annotated that it takes bytes and the CFFI message is quite clear what you did wrong if you pass a `str`. I see no reason to support `str`. Since the introduction of Python 3 a long time ago, people should be quite familiar with this distinction between `str` and `bytes` already.

tests/test_random.py
```diff
@@ -38,13 +39,45 @@ def test_bytes(rng):
...
+    """
+    Test that consecutive reseeding of the random number generator works.
+    """
+    # Using our own rng for getting random seed sizes.
```
dgarske: **🔵 [Low] Misleading comment in test_reseed_multiple (no random seed sizes used)**

The comment `# Using our own rng for getting random seed sizes.` does not match the code: the loop uses a fixed seed of size 32 on every iteration and no randomness at all. This looks like a copy/paste leftover from an earlier version of the test and is confusing to future readers.

**Fix:** Remove or correct the inaccurate comment so it describes what the test actually does.

scripts/build_ffi.py
dgarske: **🔵 [Low] HASHDRBG feature detection relies on explicit #define that wolfSSL may only set implicitly**

`features["HASHDRBG"]` is detected by searching for the literal string `#define HAVE_HASHDRBG` in `options.h`/`user_settings.h`. Unlike most features, wolfSSL auto-enables Hash-DRBG *inside* `random.h` (see the bundled header: `#if !defined(WC_NO_HASHDRBG) && !defined(CUSTOM_RAND_GENERATE_BLOCK)` -> `#define HAVE_HASHDRBG`), so the symbol is not guaranteed to appear in `options.h`/`user_settings.h`. For the bundled configure output it does appear (options.h lines 131-132), and the PR correctly adds it to `windows/non_fips/user_settings.h`, so current builds are fine. However, a user supplying a custom `user_settings.h` that relies on the implicit default (defines neither `HAVE_HASHDRBG` nor `WC_NO_HASHDRBG`) would build a library that fully supports `wc_RNG_DRBG_Reseed` yet get `HASHDRBG=0`, silently dropping the `reseed()` method. This is a false-negative (safe, no crash) and is consistent with how other flags are detected, so it is not blocking.

**Fix:** Consider mirroring wolfSSL's own gating logic (default-on unless WC_NO_HASHDRBG / CUSTOM_RAND_GENERATE_BLOCK), or document that HAVE_HASHDRBG must be explicitly defined in user_settings.h for reseed support. Low priority since configure builds and the new Windows non_fips settings both define it explicitly.

dgarske (commented):
## Skoll Code Review

**1 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Low] **Misleading comment in test_reseed_multiple (no random seed sizes used)** — `tests/test_random.py:76`
- [Low] **HASHDRBG feature detection relies on explicit #define that wolfSSL may only set implicitly** — `scripts/build_ffi.py:381`

#### Type stub lib.pyi not updated with HASHDRBG_ENABLED / wc_RNG_DRBG_Reseed
**File:** `wolfcrypt/_ffi/lib.pyi:33`
**Function:** `N/A (module-level stub)`
**Severity:** Medium

The repo maintains a hand-written CFFI type stub at `wolfcrypt/_ffi/lib.pyi` that enumerates every `*_ENABLED` feature flag (lines 7-33, including `HKDF_ENABLED`, `ML_DSA_ENABLED`, etc.). This PR introduces a new build-time flag `HASHDRBG_ENABLED` (used in `wolfcrypt/random.py:81` and `tests/test_random.py:57,71`) and a new native function `wc_RNG_DRBG_Reseed` (used in `wolfcrypt/random.py:87`), but does not add either to the stub. The project ships `mypy` and `ty` in its `[dependency-groups] dev` (see `pyproject.toml`), so a type check would report `_lib.HASHDRBG_ENABLED` as `attr-defined` error. Adding the flag keeps the stub's flag section complete and consistent.

**Recommendation:** Add `HASHDRBG_ENABLED: int` to the flags section of `lib.pyi` (and, for completeness, `wc_RNG_DRBG_Reseed`) so the stub stays in sync with the CFFI-generated symbols this PR references.

_Referenced code: `wolfcrypt/_ffi/lib.pyi:33-36` (4 lines)_

---

dgarske (changes requested):
https://github.com/wolfSSL/wolfcrypt-py/pull/130#pullrequestreview-4637849658

tests/test_random.py
mjdemilliano: Fixed

scripts/build_ffi.py
mjdemilliano: Fixed

dgarske (approved):

PR #131, opened 2026-06-26
Fix license in pyproject.toml

mjdemilliano:
This change makes pyproject.toml compliant to PEP-621 w.r.t. license specification.

Update: changed to correct reference.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,8 +6,7 @@ requires-python = ">=3.10"
 authors = [
     {name = "wolfSSL Inc.", email="info@wolfssl.com"},
 ]
-license = "GPL-2.0-only OR LicenseRef-WolfSSL"
-license-files = ["LICENSING.rst"]
+license = {file = "LICENSING.rst"}
 keywords = ["wolfssl", "wolfcrypt", "security", "cryptography"]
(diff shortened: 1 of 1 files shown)


roberthdevries:
The current specification for python package information that specifies what goes into a pyproject.toml file can be found in [pyproject.toml specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#)
[PEP 639](https://peps.python.org/pep-0639/) has been superseded by the above spec. Assuming that the above-mentioned [PEP 612](https://peps.python.org/pep-0612/) is a mistake as it discusses Parameter Specification Variables related to typing.

The proposed changes use a construct marked as legacy.

mjdemilliano:
You're right. The number should have been PEP-621 and has indeed been superseded. Discussion to be held about which PEP this repository still has to support.

sebastian-carpenter:
@mjdemilliano do you need support for the legacy version? The current code seems to reflect the standard and I'm not sure there's an easy way to support both.

mjdemilliano:
I currently still need it but doesn't have to be upstream, so if you are ok with not supporting the legacy version then let's not change it back.

PR #133, opened 2026-06-28
Change workflow to using uv and ruff.

roberthdevries:
uv is used to manage virtual environments and building wheels and source distributions.
This replaces tox.
ruff is used to do static checking.
README is updated to reflect this change.


dgarske (commented):
## Skoll Code Review

**7 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **build-no-pqc 'Install uv' step has both uses and run (invalid workflow)** — `.github/workflows/python-app.yml:79-87`
- [High] **'Install wolfcrypt-py' step left with only env and loses USE_LOCAL_WOLFSSL** — `.github/workflows/python-app.yml:76-78`
- [High] **uv sync --locked will fail: no uv.lock committed in the repo** — `.github/workflows/python-app.yml:32`
- [Medium] **README documents invalid uv run sync --dev command** — `README.rst:93`
- [Medium] **README uv pip install . requires an active virtualenv** — `README.rst:67`
- [Low] **RST title underline too short for renamed 'pytest' heading** — `README.rst:83-84`
- [Low] **--all-extras is a no-op (no optional-dependencies defined)** — `.github/workflows/python-app.yml:32`

.github/workflows/python-app.yml
```diff
@@ -80,13 +76,18 @@ jobs:
     - name: Install wolfcrypt-py against the local wolfSSL
       env:
         USE_LOCAL_WOLFSSL: ${{ github.workspace }}/wolfssl-install
+    - name: Install uv
```
dgarske: **🔴 [High] build-no-pqc 'Install uv' step has both uses and run (invalid workflow)**

The new `Install uv` step was inserted in the MIDDLE of the existing `Install wolfcrypt-py against the local wolfSSL` step. As a result the `Install uv` step now carries BOTH a `uses:` (astral-sh/setup-uv) AND a `run:` block (`uv venv` / `uv pip install ...`). GitHub Actions forbids a step from having both `uses` and `run`; the workflow file is rejected at validation time, so the ENTIRE `build-no-pqc` job (the regression coverage for issue #2659) never runs. Confirmed by parsing the YAML: `build-no-pqc step[5]` has keys `['name','uses','with','run']`.

**Fix:** Move the `Install uv` step ABOVE the install step so each step has exactly one of `uses`/`run`. Keep the `uv venv`/`uv pip install` commands under the install step (with its `env`).

.github/workflows/python-app.yml
```diff
@@ -80,13 +76,18 @@ jobs:
     - name: Install wolfcrypt-py against the local wolfSSL
```
dgarske, line 87: **🔴 [High] 'Install wolfcrypt-py' step left with only env and loses USE_LOCAL_WOLFSSL**

After the botched insertion, the `Install wolfcrypt-py against the local wolfSSL` step has only `name` and `env` and NO `run`/`uses` (confirmed: `build-no-pqc step[4]` keys `['name','env']`) — an invalid/empty step. Critically, even if the both-keys problem in the following step is fixed, the `USE_LOCAL_WOLFSSL` environment variable is now stranded on this empty step and is NOT applied to the actual `uv pip install -e .` commands, which live on the separate `Install uv` step. The purpose of `build-no-pqc` (issue #2659) is to build wolfcrypt-py against the locally built no-PQC wolfSSL; without `USE_LOCAL_WOLFSSL` in scope of the install commands, the package is no longer built against that local wolfSSL and the regression no longer tests what it claims to.

**Fix:** Re-attach the `run:` install commands to the step that carries `env: USE_LOCAL_WOLFSSL`, and put the `Install uv` action as its own separate preceding step.

.github/workflows/python-app.yml
```diff
@@ -23,21 +23,17 @@ jobs:
...
-    - name: Test with tox
-      run: |
-        tox
+    - name: Install uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+      with:
+        # Install a specific version of uv.
+        version: "0.11.25"
+    - name: Install the project
+      run: uv sync --locked --all-extras --dev
```
dgarske: **🔴 [High] uv sync --locked will fail: no uv.lock committed in the repo**

The `build` job runs `uv sync --locked --all-extras --dev`. The `--locked` flag asserts that an up-to-date `uv.lock` already exists and must remain unchanged; if the lockfile is missing or stale, uv exits with an error instead of generating it. There is no `uv.lock` in the repository (confirmed: file absent in the working tree and not part of this PR / not tracked by git), so the fresh CI checkout has no lockfile and this step will fail immediately. The main `build` job — which now runs the tests and the new `ruff` static checks — will therefore never succeed.

**Fix:** Either commit a `uv.lock` (run `uv lock` and add it to the PR) so `--locked` is satisfiable, or drop the `--locked` flag so uv can resolve/create the lockfile on the fly.

README.rst
```diff
@@ -82,25 +80,41 @@ Testing
...
-To run the unit tests in the source code, you'll need ``tox`` and a few other
+To run the unit tests in the source code, you'll need ``uv`` and a few other
 requirements.
 
 1. Make sure that the testing requirements are installed:
 
 .. code-block:: console
 
-    $ sudo -H pip install -r requirements/test.txt
+    $ uv run sync --dev
```
dgarske: **🟠 [Medium] README documents invalid uv run sync --dev command**

Step 1 instructs `$ uv run sync --dev`. `uv run \<cmd>` executes a command/entrypoint named `\<cmd>` inside the project environment, so `uv run sync` tries to run a (non-existent) executable called `sync` and fails. The intended command to install the dev dependency group is `uv sync --dev` (matching the workflow's `uv sync ... --dev`). The surrounding prose ('Make sure that the testing requirements are installed') confirms the intent is `uv sync`.

**Fix:** Change `uv run sync --dev` to `uv sync --dev`.

README.rst
```diff
@@ -66,7 +64,7 @@ should be set equal to the installation path for the wolfSSL library:
 
 .. code-block:: bash
 
-    $ USE_LOCAL_WOLFSSL=/path/to/wolfssl/install pip install .
+    $ USE_LOCAL_WOLFSSL=/path/to/wolfssl/install uv pip install .
```
dgarske: **🟠 [Medium] README uv pip install . requires an active virtualenv**

The Installation section was changed from `pip install .` to `USE_LOCAL_WOLFSSL=/path/to/wolfssl/install uv pip install .`. Unlike plain `pip`, `uv pip install` requires an active/discoverable virtual environment (or `--system`); with no `.venv` present it errors with 'No virtual environment found...'. The CI's `build-no-pqc` job works only because it runs `uv venv` first, but the README gives the user no such step, so the documented command fails out-of-the-box for a fresh user. The previous `pip install .` worked without a venv, so this is a regression in the docs.

**Fix:** Add a `uv venv` prerequisite, use `--system`, or keep `pip install .` for the end-user installation instructions.

README.rst
```diff
@@ -82,25 +80,41 @@ Testing
     >>> Sha256('wolfcrypt').hexdigest()
     b'96e02e7b1cbcd6f104fe1fdb4652027a5505b68652b70095c6318f9dce0d1844'
 
-Testing ``wolfcrypt``'s source code with ``tox``
+Testing ``wolfcrypt``'s source code with ``pytest``
```
dgarske, line 94: **🔵 [Low] RST title underline too short for renamed 'pytest' heading**

The heading text was lengthened from '...with ``tox``' (48 chars) to '...with ``pytest``' (51 chars), but the `---` underline below it was left at 48 characters (confirmed: title length 51, underline length 48). reStructuredText requires the underline to be at least as long as the title; docutils/sphinx will emit a 'Title underline too short' warning, which can fail doc builds run with `-W`.

**Fix:** Extend the underline to at least 51 dashes to match the new title length.

.github/workflows/python-app.yml
dgarske: **🔵 [Low] --all-extras is a no-op (no optional-dependencies defined)**

`uv sync --locked --all-extras --dev` passes `--all-extras`, which enables all groups under `[project.optional-dependencies]`. The pyproject.toml defines no `[project.optional-dependencies]` (only a `[dependency-groups] dev` group, already covered by `--dev`), so `--all-extras` does nothing. Harmless, but misleading to future maintainers who may assume extras exist.

**Fix:** Drop `--all-extras` unless/until `[project.optional-dependencies]` are added.

roberthdevries: Fixed, removed the --sync flag and --all-extras flag.

roberthdevries: Removed.

.github/workflows/python-app.yml
roberthdevries: Fixed.

.github/workflows/python-app.yml
roberthdevries: Fixed.

README.rst
roberthdevries: With uv it is even simpler, you can just run `uv sync` and this will prepare the venv and build wolfcrypt-py
So the suggested command is updated to reflect that workflow.

README.rst
roberthdevries: Fixed

.github/workflows/python-app.yml
```diff
@@ -81,12 +82,12 @@ jobs:
       env:
         USE_LOCAL_WOLFSSL: ${{ github.workspace }}/wolfssl-install
       run: |
```
dgarske, line 88: **🟠 [Medium] build-no-pqc mixes uv pip install with uv run, risking an extension rebuild without USE_LOCAL_WOLFSSL**

The install step uses uv's `pip` interface (`uv venv` + `uv pip install -e .`) with `USE_LOCAL_WOLFSSL` set so the CFFI extension builds against the locally-built no-PQC wolfSSL. The next two steps switch to uv's `project` interface (`uv run python -c ...`, `uv run pytest tests`). In a directory containing a `[project]` table, `uv run` performs an automatic project sync (lock + exact sync) before executing the command. If that sync rebuilds/reinstalls the `wolfcrypt` extension, the `Import smoke` step runs `uv run` with NO `USE_LOCAL_WOLFSSL` in its env, so `scripts/build_ffi.py` falls back to building wolfSSL from the `lib/wolfssl` submodule with default features (which include ML-KEM/ML-DSA). That would declare `INVALID_DEVID` and silently defeat the exact regression scenario this job exists to cover (issue #2659), or fail the build. Mixing the pip and project interfaces in one job is the documented footgun here.

**Fix:** Verify uv does not rebuild the extension between steps; if it can, add `--no-sync` to the `uv run` invocations and set `USE_LOCAL_WOLFSSL` on the smoke step (or convert the whole job to `USE_LOCAL_WOLFSSL=... uv sync` up front) so the regression coverage stays valid.

README.rst
```diff
@@ -75,32 +73,50 @@ features based on how native wolfSSL has been compiled.  It uses the
 
 Testing
 -------
+.. code-block:: console
```
dgarske, line 84: **🟠 [Medium] Malformed reStructuredText: code-block directive missing blank line**

The newly added console block has no blank line between the directive header and its body. RST requires a blank line after a directive before its content; without it docutils raises `Error in "code-block" directive` and the snippet will not render as a code sample on PyPI/GitHub (the long_description is used as the PyPI page via setup.py). All the other code blocks in this file correctly include the blank line.

**Fix:** Insert a blank line between `.. code-block:: console` and `$ uv run python3`.

.github/workflows/python-app.yml
```diff
@@ -23,21 +23,17 @@ jobs:
...
-      run: |
-        tox
+    - name: Install uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+      with:
+        # Install a specific version of uv.
+        version: "0.11.25"
+    - name: Install the project
+      run: uv sync --dev
+    - name: Perform static checks
```
dgarske, line 35: **🟠 [Medium] Enabling uv run ruff check may fail CI on previously-unlinted code**

The PR removes the commented-out flake8 block that carried the note `# Won't pass flake8 yet` and replaces it with an ACTIVE `uv run ruff check` step (configured rules `E4,E7,E9,F,B,UP`, over `wolfcrypt/*.py` and `scripts/build_ffi.py`). If that code has not been cleaned up to satisfy those rules, the `build` job now fails on every push/PR. The README shows passing pytest output but no ruff output, so ruff cleanliness is unverified in the diff.

**Fix:** Run `uv run ruff check` against the current tree and fix (or narrow the rule set / add `# noqa`) any violations before merging, so CI stays green.

README.rst
```diff
@@ -66,7 +64,7 @@ should be set equal to the installation path for the wolfSSL library:
 
 .. code-block:: bash
 
-    $ USE_LOCAL_WOLFSSL=/path/to/wolfssl/install pip install .
+    $ USE_LOCAL_WOLFSSL=/path/to/wolfssl/install uv sync
```
dgarske: **🔵 [Low] README tells users to uv sync to install against local wolfSSL, changing install semantics**

The Installation section replaced `USE_LOCAL_WOLFSSL=/path/... pip install .` with `USE_LOCAL_WOLFSSL=/path/... uv sync`. `uv sync` installs the project into a project-local `.venv` and only works from within the wolfcrypt-py source tree, whereas `pip install .` installed wolfcrypt into the user's active environment. A user following these Installation instructions to consume wolfcrypt-py will not get it on their normal Python path (they would have to `uv run`), which is a behavior change from the documented intent of this section.

**Fix:** Use `uv pip install .` here, or clarify that the package is installed into the project `.venv` and must be run via `uv run`.

pyproject.toml
```diff
@@ -40,21 +40,10 @@ dev = [
     "ruff",
     "sphinx",
     "sphinx-rtd-theme",
-    "tox >= 4",
     "ty",
```
dgarske, line 44: **🔵 [Low] tox tested a built wheel (package = "wheel"); uv run pytest tests the editable/source install**

The removed `[tool.tox.env_run_base]` used `package = "wheel"`, so tests ran against a freshly built wheel and thus exercised packaging (MANIFEST / `package_data` such as the bundled `*.dll` and `*.pyi`). The new flow runs `uv run pytest` against an editable/source install, which can miss packaging regressions (files that are present in the tree but not packaged into the wheel). This is a coverage change worth being deliberate about.

**Fix:** If packaging coverage matters, add a job/step that builds the wheel (`uv build --wheel`) and runs the tests against the installed wheel, in addition to the editable-install run.

---
> **Note:** Referenced line (`pyproject.toml:43-51 (removed [tool.tox])`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

**Findings:** 6 total — 6 posted, 0 skipped
**5 finding(s) posted as inline comments** (see file-level comments below)
**1 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Medium] **build-no-pqc mixes uv pip install with uv run, risking an extension rebuild without USE_LOCAL_WOLFSSL** — `.github/workflows/python-app.yml:84-93`
- [Medium] **Malformed reStructuredText: code-block directive missing blank line** — `README.rst:76-77`
- [Medium] **Enabling uv run ruff check may fail CI on previously-unlinted code** — `.github/workflows/python-app.yml:33-34`
- [Low] **README tells users to uv sync to install against local wolfSSL, changing install semantics** — `README.rst:67`
- [Low] **tox tested a built wheel (package = "wheel"); uv run pytest tests the editable/source install** — `pyproject.toml:43-51 (removed [tool.tox])`

#### Incomplete tox removal: Makefile test-all/check-all still invoke tox
**File:** `Makefile:61-64`
**Function:** `test-all / check-all`
**Severity:** Low

This PR removes tox from `pyproject.toml` and `requirements/test.txt`, but `make test-all` and `make check-all` still call `tox`. Because tox is no longer a declared test dependency, `uv sync` / `pip install -r requirements/test.txt` will not install it and those Makefile targets now break. This is a direct consequence of the diff (the Makefile is outside the changed-file set but the tox removal makes it dead/broken).

**Recommendation:** Update or remove the tox-based Makefile targets to keep the tox->uv migration consistent.

_Referenced code: `Makefile:61-64` (4 lines)_

---

dgarske (changes requested):
https://github.com/wolfSSL/wolfcrypt-py/pull/133#pullrequestreview-4621775199

.github/workflows/python-app.yml
roberthdevries: The static check passes in the pipeline verification and prints a satisfying `All checks passed!` message.

.github/workflows/python-app.yml
roberthdevries: Added a step to verify the no-pqc doesn't have ML-DSA and ML-KEM enabled.

pyproject.toml
roberthdevries: The build step now explicitly builds a wheel. This wheel is subsequently used to run the tests, replacing the source tree install.

README.rst
roberthdevries: Fixed.

README.rst
roberthdevries: Expanded the explanation to show how to build and install the wheel using uv.

Also fixed any remaining references to pip in places where the follow up mentions uv.

Makefile
```diff
@@ -49,19 +49,21 @@ clean-test: ## remove test and coverage artifacts
...
+test-all: ## run tests on every Python version with uv
+	for version in 3.10 3.11 3.12 3.13 3.14; do \
+		echo "=== Python $$version ==="; \
+		uv run --python $$version pytest; \
+	done
 
-check-all: test-all ## run tests on every Python version with tox
+check-all: test-all ## run tests on every Python version with uv
 
 cov: install ## check code coverage quickly with the default Python
```
dgarske: **🔴 [High] Makefile targets depend on removed install target**

This PR deletes the `install` target (`install: clean` / `python setup.py install`) at the bottom of the Makefile, but three targets still declare `install` as a prerequisite: `cov: install` (line 68), `docs: install` (line 74), and `doctest: install` (line 79). `servedocs` depends on `docs`, so it is transitively affected too. With no `install` rule defined, running `make cov`, `make docs`, `make doctest`, or `make servedocs` now fails immediately with `make: *** No rule to make target 'install'. Stop.` This is a regression in developer tooling introduced by the diff.

**Fix:** Either drop the now-broken `install` prerequisite from `cov`, `docs`, and `doctest`, or reintroduce an `install` target implemented with uv (e.g. `uv pip install .`) so these targets keep working. Do not leave a dangling prerequisite.

Makefile
```diff
@@ -49,19 +49,21 @@ clean-test: ## remove test and coverage artifacts
...
-test: install ## run tests quickly with the default Python
-	py.test tests
+test: ## run tests quickly with the default Python
+	uv run py.test tests
 
 check: test ## run tests quickly with the default Python
 
-test-all: ## run tests on every Python version with tox
-	tox
+test-all: ## run tests on every Python version with uv
```
dgarske, line 60: **🟠 [Medium] test-all loop hides per-version test failures and omits the test path**

The new `test-all` recipe runs a shell `for` loop over Python versions. Two problems: (1) A failing `uv run ... pytest` inside the loop does not stop iteration, and the recipe's exit status is that of the final iteration only. If Python 3.10 fails but 3.14 passes, `make test-all` (and therefore `make check-all`) reports success, masking real failures. (2) Unlike the `test` target (`uv run py.test tests`) and the CI workflow (`uv run pytest tests`), this invocation calls `pytest` with no path argument. Since there is no `testpaths`/`[tool.pytest.ini_options]` configuration in `pyproject.toml`, pytest falls back to default recursive discovery from the repo root, which can pick up unintended files (e.g. under the `lib/` submodule) or behave inconsistently with the other targets.

**Fix:** Make the loop fail fast (`|| exit 1`, or `set -e`) so any failing version fails the target, and pass the `tests` path explicitly to match the other targets and the CI workflow.

.github/workflows/python-app.yml
```diff
@@ -81,12 +86,16 @@ jobs:
       env:
         USE_LOCAL_WOLFSSL: ${{ github.workspace }}/wolfssl-install
       run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements/test.txt
-        pip install -e .
+        uv venv
+        uv pip install -r requirements/test.txt
+        uv pip install -e .
+    - name: Check absence of ML-KEM and ML-DSA support
```
dgarske, line 92: **🟠 [Medium] uv run auto-sync may rebuild wolfcrypt without USE_LOCAL_WOLFSSL in the no-PQC regression checks**

The no-PQC job installs wolfcrypt editable against the locally built (ML-KEM/ML-DSA-disabled) wolfSSL with `uv pip install -e .` while `USE_LOCAL_WOLFSSL` is set. It then runs the regression assertions via `uv run python -c "..."` and `uv run pytest tests`. By default `uv run` performs an automatic project sync before executing, which can rebuild/reinstall the `wolfcrypt` project from source. The 'Check absence of ML-KEM and ML-DSA support' and 'Import smoke' steps do NOT set `USE_LOCAL_WOLFSSL` in their `env:`, so if `uv run` re-syncs it may rebuild wolfcrypt against the bundled/default wolfSSL (which enables ML-KEM/ML-DSA) or fail the build — either way undermining the very regression this job exists to protect (issue #2659). This risk is introduced by the switch from a plain `python -c`/`pytest` (which used the already-installed editable package) to `uv run`.

**Fix:** Pass `--no-sync` to the `uv run` invocations in this job (so uv uses the exact editable install produced in the previous step), and/or set `USE_LOCAL_WOLFSSL` in the `env:` of the check/smoke steps so any rebuild still targets the local no-PQC wolfSSL. Please confirm uv's re-sync behavior here so the regression is not silently invalidated.

.github/workflows/python-app.yml
```diff
@@ -18,26 +18,26 @@ jobs:
...
-#        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-#        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-#        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-    - name: Test with tox
+    - name: Install uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+      with:
+        # Install a specific version of uv.
+        version: "0.11.26"
+    - name: Build the wheel
```
dgarske, line 31: **🟠 [Medium] "Run tests using the locally built wheel" may not actually exercise the wheel**

The `build` job builds a wheel (`uv build --wheel`), then runs `uv sync --dev` (which installs the `wolfcrypt` project from source into the environment), then `uv pip install dist/*.whl`, then `uv run pytest tests`. Two things weaken the stated intent of testing the built wheel: (1) after `uv sync --dev` has already installed `wolfcrypt` at the same version, `uv pip install dist/*.whl` will typically see the requirement as already satisfied and skip actually installing the wheel; (2) `uv run pytest` auto-syncs the project environment again before running, which can reinstall from source. As a result the tests may run against the source/editable build rather than the artifact in `dist/`, so a packaging defect in the wheel could go undetected despite the step name.

**Fix:** Force-install the artifact (`uv pip install --reinstall dist/*.whl`) and run pytest with `uv run --no-sync` so the wheel actually under test is the one exercised. Alternatively, install and test the wheel in a clean venv separate from the `uv sync` project environment.

README.rst
```diff
@@ -21,32 +21,32 @@ Install the following on Windows:
...
 
-   python setup.py sdist
+   uv build --sdist
 
 
 Linux
 ^^^^^
 
-The `setup.py` file covers most things you will need to do to build and install from source. As pre-requisites you will need to install either from your OS repository or pip. You'll also need the Python development package for your Python version:
+The `setup.py` file covers most things you will need to do to build and install from source. As pre-requisites you will need to install either from your OS repository or uv. You'll also need the Python development package for your Python version:
```
dgarske: **🔵 [Low] README grammar: 'from your OS repository or uv'**

The Linux prerequisites sentence was mechanically changed from 'or pip' to 'or uv', producing awkward phrasing: 'you will need to install either from your OS repository or uv.' Reads as installing 'from uv'. Minor wording cleanup for the docs touched by this PR.

**Fix:** Reword to 'install ... from your OS repository or with uv' for clarity. Non-blocking.

dgarske (commented):
## Skoll Code Review

**Findings:** 5 total — 5 posted, 0 skipped
**5 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [High] **Makefile targets depend on removed install target** — `Makefile:68,74,79`
- [Medium] **test-all loop hides per-version test failures and omits the test path** — `Makefile:60-64`
- [Medium] **uv run auto-sync may rebuild wolfcrypt without USE_LOCAL_WOLFSSL in the no-PQC regression checks** — `.github/workflows/python-app.yml:92-101`
- [Medium] **"Run tests using the locally built wheel" may not actually exercise the wheel** — `.github/workflows/python-app.yml:31-40`
- [Low] **README grammar: 'from your OS repository or uv'** — `README.rst:45`

dgarske (changes requested):
https://github.com/wolfSSL/wolfcrypt-py/pull/133#pullrequestreview-4637763895

.github/workflows/python-app.yml
roberthdevries: Applied first suggestion

.github/workflows/python-app.yml
roberthdevries: Applied first suggestion.

Makefile
roberthdevries: Applied first suggestion and added the tests path explicitly.

Makefile
roberthdevries: Removed the obsolete install dependencies.

Fixed targets docs and doctest. (doctest fails BTW).

README.rst
roberthdevries: Fixed

dgarske (commented):
## Skoll Code Review

### Posted findings
- [Medium] **no-pqc 'Import smoke' step omits --no-sync and USE_LOCAL_WOLFSSL, undermining the regression test** — `.github/workflows/python-app.yml:96-97`
- [Low] **no-pqc 'Run tests' step omits --no-sync, causing a redundant project re-sync/rebuild** — `.github/workflows/python-app.yml:98-101`
- [Low] **Makefile 'dist' target still uses setup.py while README migrates to 'uv build'** — `Makefile:86-93`

.github/workflows/python-app.yml
```diff
@@ -81,12 +86,16 @@ jobs:
...
-        pip install -r requirements/test.txt
-        pip install -e .
+        uv venv
+        uv pip install -r requirements/test.txt
+        uv pip install -e .
+    - name: Check absence of ML-KEM and ML-DSA support
+      run: |
+        uv run --no-sync python -c "from wolfcrypt._ffi import lib as _lib; assert _lib.ML_KEM_ENABLED == 0, 'ML-KEM should be disabled'"
+        uv run --no-sync python -c "from wolfcrypt._ffi import lib as _lib; assert _lib.ML_DSA_ENABLED == 0, 'ML-DSA should be disabled'"
     - name: Import smoke (regression for INVALID_DEVID)
```
dgarske, line 96: **🟠 [Medium] no-pqc 'Import smoke' step omits --no-sync and USE_LOCAL_WOLFSSL, undermining the regression test**

The `build-no-pqc` job builds a no-PQC wolfSSL, installs wolfcrypt against it via a manually-managed venv (`uv venv` + `uv pip install -r requirements/test.txt` + `uv pip install -e .` with `USE_LOCAL_WOLFSSL` set), and the immediately-preceding 'Check absence of ML-KEM and ML-DSA support' step correctly runs with `uv run --no-sync ...`. The new 'Import smoke' step instead runs `uv run python -c "..."` WITHOUT `--no-sync` AND without setting `USE_LOCAL_WOLFSSL`. Because there is no committed `uv.lock` and the repo is a uv project (pyproject.toml has `[project]`), a plain `uv run` operates in project mode and will re-sync the environment first, reinstalling/rebuilding the `wolfcrypt` project from source. With `USE_LOCAL_WOLFSSL` unset, `scripts/build_ffi.py` falls back to building wolfSSL from the `lib/wolfssl` submodule with its default flags, which include `--enable-mlkem` and `--enable-mldsa` (confirmed in `make_flags()`). That rebuild would replace the carefully-built no-PQC wolfcrypt with a PQC-enabled one, so this regression test for issue #2659 (INVALID_DEVID only declared when ML_KEM/ML_DSA enabled) would silently validate the wrong configuration -- or fail/hang rebuilding wolfSSL. The inconsistency with the adjacent `--no-sync` step strongly suggests this is an oversight.

**Fix:** Add `--no-sync` (and, defensively, the `USE_LOCAL_WOLFSSL` env) to the 'Import smoke' step so it exercises the already-installed no-PQC build instead of triggering a project re-sync/rebuild against a default PQC-enabled wolfSSL.

.github/workflows/python-app.yml
```diff
@@ -81,12 +86,16 @@ jobs:
...
-      run: python -c "from wolfcrypt.random import Random; Random()"
+      run: uv run python -c "from wolfcrypt.random import Random; Random()"
     - name: Run tests
```
dgarske, line 98: **🔵 [Low] no-pqc 'Run tests' step omits --no-sync, causing a redundant project re-sync/rebuild**

The final 'Run tests' step runs `uv run pytest tests` without `--no-sync`, unlike the 'Check absence' step which uses `--no-sync`. Since the venv was populated manually with `uv pip install`, a plain `uv run` will re-sync the project environment before running, discarding the manual `uv pip install -e .` result and rebuilding wolfcrypt. Here `USE_LOCAL_WOLFSSL` is set on the step, so the rebuilt target is still the no-PQC library and the test remains correct -- but the rebuild is redundant work and is inconsistent with the surrounding steps' use of `--no-sync`.

**Fix:** Use `uv run --no-sync pytest tests` for consistency with the other steps in this manually-managed-venv job and to avoid an unnecessary rebuild.

Makefile
```diff
@@ -49,34 +49,36 @@ clean-test: ## remove test and coverage artifacts
...
 	$(BROWSER) docs/_build/html/index.html
 
-doctest: install ## generate Sphinx HTML documentation, including API docs
+doctest: ## generate Sphinx HTML documentation, including API docs
 	$(MAKE) -C docs clean
-	$(MAKE) -C docs doctest
+	uv run $(MAKE) -C docs doctest
 
 servedocs: docs ## compile the docs watching for changes
 	watchmedo shell-command -p '*.rst' -c '$(MAKE) -C docs html' -R -D .
```
dgarske, line 84: **🔵 [Low] Makefile 'dist' target still uses setup.py while README migrates to 'uv build'**

This PR migrates the documented build workflow to `uv build --sdist` / `uv build --wheel` (README.rst) and removes the `install` target that shelled out to `python setup.py install`. However the `dist` target (unchanged by the diff) still runs `python setup.py sdist` directly. This is not a correctness bug and the line is outside the direct diff, but it leaves the Makefile inconsistent with the new uv-based direction the PR establishes. Worth aligning while touching this file.

**Fix:** Optionally switch the `dist` target's sdist build to `uv build --sdist` for consistency with the README and the rest of the migration.

.github/workflows/python-app.yml
roberthdevries: Added `--no-sync` everywhere and moved the `env` setting to the top of the jobs section so that it applies to all steps in the no-pqc job. This ensures that the environment is in every step.

.github/workflows/python-app.yml
roberthdevries: Fixed.

Makefile
roberthdevries: Switched to `uv build --sdist`

Please note that the subsequent steps refer to scripts in a non-existent `make` sub-directory. Not sure where these files went.
Also twine is not installed as a dependency. Note that `uv` is capable of uploading the distribution to pypi.

roberthdevries:
The last patch also removes the now obsolete `requirements` subdirectory. All information is now present in the `pyproject.toml` file as single source of truth.

dgarske (approved):

PR #134, opened 2026-06-30
Prepare for v5.9.2 release

mjdemilliano:


sebastian-carpenter (dismissed):
One minor nitpick:

  INFO-3: Typo "ML-DSDA" in comment [NIT] (style)
  File:       scripts/build_ffi.py:382
  Function:   get_features
  Confidence: High

  The comment reads "Determine if support for ML-DSDA signing & verification without context has been enabled." — "ML-DSDA" should be "ML-DSA".

  Code:
    # Determine if support for ML-DSDA signing & verification without context has been enabled.

  Recommendation: Fix the typo.

Besides that the dilithium function signatures should be updated to the mldsa ones in the future. It looks like macro's reassign them right now so it shouldn't be blocking.

wolfcrypt/ciphers.py
```diff
@@ -2440,6 +2449,8 @@ def sign(self, message, rng=None, ctx=None):
                 )
                 if ret < 0:  # pragma: no cover
                     raise WolfCryptApiError("wc_dilithium_sign_ctx_msg() error", ret)
+            elif not _lib.ML_DSA_NO_CTX_ENABLED:
+                raise WolfCryptError("support for signing without context is disabled")
```
dgarske: **🔵 [Low] sign() constructs a Random() before the disabled-support check that raises**

When `ML_DSA_NO_CTX_ENABLED` is 0 and `sign(message)` is called with no `rng` and `ctx=None`, the method first seeds a fresh `Random()` (`if rng is None: rng = Random()`) and only afterward reaches the `elif not _lib.ML_DSA_NO_CTX_ENABLED: raise WolfCryptError(...)` branch. The RNG allocation/seeding is wasted work on a path that always raises. Minor, and not a correctness issue.

**Fix:** Optional cleanup: raise the not-supported error before allocating the RNG. Not required for merge.

dgarske (commented):
## Skoll Code Review

**2 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Low] **sign() constructs a Random() before the disabled-support check that raises** — `wolfcrypt/ciphers.py:2428-2453`

#### MIN_AUTH_TAG_SZ default of 12 rejects short GCM tags the underlying build actually accepts
**File:** `scripts/build_ffi.py:376-378, wolfcrypt/ciphers.py:407-411`
**Function:** `get_features / AesGcmStream.__init__`
**Severity:** Medium

When `WOLFSSL_MIN_AUTH_TAG_SZ` is not emitted in `options.h`/`user_settings.h`, `features["MIN_AUTH_TAG_SZ"]` falls back to `12`. wolfSSL only enforces a GCM minimum tag length when that macro is explicitly defined (the streaming encrypt/`final` path otherwise accepts any 4-16 byte tag), so a default wolfSSL build imposes no minimum. The previous release's `test_invalid_tag_bytes` asserted that `tag_bytes` of 4 and 8 succeed on the default CI build, confirming the C library accepts them. With this change, `AesGcmStream(key, iv, tag_bytes=4)` and `tag_bytes=8` now raise `ValueError` on the same default build. This is a behavior change from 5.8.4 that the ChangeLog does not mention. The value 12 (96-bit tag) is a defensible NIST SP 800-38D general-use floor, so this may be intentional hardening -- but the Python default is stricter than the actual library configuration, and callers using 4/8-byte tags will break silently on upgrade.

**Recommendation:** Confirm the default of 12 is intended. If the goal is to mirror the build, default to 0 so the binding matches an unconstrained wolfSSL build. If the stricter floor is intentional, add a ChangeLog entry documenting that GCM authentication tags shorter than 12 bytes are now rejected by default.

_Referenced code: `scripts/build_ffi.py:376-378, wolfcrypt/ciphers.py:407-413` (7 lines)_

---

#### No-context ML-DSA sign/verify now disabled by default is a breaking change not called out in ChangeLog
**File:** `scripts/build_ffi.py:236, wolfcrypt/ciphers.py:2452-2453,2273-2274,2519-2520`
**Function:** `make_flags / MlDsaPrivate.sign / _MlDsaBase.verify / MlDsaPrivate.sign_with_seed`
**Severity:** Medium

The build now uses `--enable-mldsa=yes` (without `no-ctx`), so `ML_DSA_NO_CTX_ENABLED` is 0 in default builds. As a result `mldsa_priv.sign(message)`, `mldsa_pub.verify(sig, msg)`, and `sign_with_seed(msg, seed)` -- all with `ctx=None` -- now raise `WolfCryptError("support for ... without context is disabled")` on a default build. In 5.8.4 (`--enable-dilithium`, which enabled the no-context variant), these no-context calls worked. This is a user-visible API behavior change: code that previously signed/verified without a context will now raise. The ChangeLog lists "ML-DSA: Add support for signing and verifying with context" but does not state that no-context operation is now opt-in (requires building with `yes,no-ctx`). Aligning the default with FIPS-204 final (which mandates a context) is reasonable, but the migration impact should be documented.

**Recommendation:** Document the default-off no-context behavior in the ChangeLog so users upgrading from 5.8.4 understand why `sign(message)`/`verify(sig, msg)` without a context now raise.

_Referenced code: `scripts/build_ffi.py:236, wolfcrypt/ciphers.py:2452-2453,2273-2274,2519-2520` (4 lines)_

---

wolfcrypt/ciphers.py
mjdemilliano: Done. Moved the if statement up, and did the same for the other routines so that this is consistent.

mjdemilliano:
About the `MIN_AUTH_TAG_SZ `: setting the default to 0 in the `build_ffi.py` script break the test suite, because the default of 12 cannot be read from `options.h`, which is the only file that is considered by the feature detection part. So keeping the default of 12 in the Python library. I think this only has implications for people running the test suite with a custom wolfSSL build, which I don't believe we have to make provisions for. So I will keep this as it is.

The comment about the changelog has been processed.

dgarske (approved):

sebastian-carpenter (approved):

PR #137, opened 2026-07-08
Clean up setup.py

roberthdevries:
- remove obsolete merging of LICENSING.rst in description as the include is not there anymore
- put setuptools configuration in pyproject.toml
- the setup_requires setting in setup.py is deprecated and is covered by pyproject.toml [build-system] - requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@
 import os
 import re
 import sys
-from setuptools import setup, find_packages
+from setuptools import setup
 
 os.chdir(os.path.dirname(sys.argv[0]) or ".")
 
@@ -47,20 +47,11 @@ else:
 with open("README.rst") as readme_file:
     long_description = readme_file.read()
 
-with open("LICENSING.rst") as licensing_file:
-    long_description = long_description.replace(".. include:: LICENSING.rst\n",
-                                                licensing_file.read())
-
 setup(
     name="wolfcrypt",
     version=verstr,
     long_description=long_description,
     long_description_content_type='text/x-rst',
 
-    packages=find_packages(),
-
-    setup_requires=["cffi>=1.17"],
     cffi_modules=["./scripts/build_ffi.py:ffibuilder"],
-
-    package_data={"wolfcrypt": ["*.dll", "**/*.pyi", "py.typed"]},
 )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,9 @@ dev = [
     "types-cffi",
 ]
 
+[tool.setuptools.packages.find]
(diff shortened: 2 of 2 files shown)


dgarske (changes requested):
Please resolve merge conflicts. Thanks!

roberthdevries:
Rebased on master, conflicts are resolved.

dgarske (approved):

setup.py
```diff
@@ -47,20 +47,11 @@
...
     version=verstr,
     long_description=long_description,
     long_description_content_type='text/x-rst',
 
-    packages=find_packages(),
-
-    setup_requires=["cffi>=1.17"],
     cffi_modules=["./scripts/build_ffi.py:ffibuilder"],
-
-    package_data={"wolfcrypt": ["*.dll", "**/*.pyi", "py.typed"]},
```
sebastian-carpenter, line 65: it looks like '*.dll' is being dropped from the build. The '*.pyi' and 'py.typed' should be fine to remove since I believe they're implicit but a rule needs to be added to pyproject.toml to continue copying '*.dll'.

sebastian-carpenter (changes requested):

setup.py
roberthdevries: The `include` line on line 50 of `pyproject.toml` specifies the directories to include in the wheel. On Linux it includes the .so the file by default. So my hopes are that on Windows setuptools would also do this for .dll files by default.
Unfortunately I do not have a development environment under Windows available. Could someone at your end try to run `uv build --wheel` on a Windows system?

sebastian-carpenter: That should be doable. I'll get back to you once I've tested it.

roberthdevries: I tried adding a workflow to the github actions, but it seems to crash in one of the tests. In any case the tests seems to run for a bit, so at least something is being built. It only seems that no .dll is built, but an .exp file. In any case that file is not included in the wheel (it is not using the changes of this patch).

sebastian-carpenter: Ahh, so this may be broken already - unless we're misunderstanding how this is supposed to be built. Debugging workflows is a bit of a chore so I'll see if I can get it working locally.

I've got a desktop with Windows that I'll get setup this weekend, we'll see what I find come Monday.

sebastian-carpenter: Hey @roberthdevries, I haven't had as much time as I would have liked to investigate this. I've tried several methods to get Windows building wolfcrypt-py but I seem to be doing something incorrectly. It did seem like I got it working one time but the .dll wasn't copied over at all, so it could be that something is already broken.

I'll leave this with you for now, if you can get that workflow working for Windows that would be much appreciated.

